PR #13, opened 2021-08-30
aiohttp HTTP client

Lunarmagpie:
### Changes

- `adds`: Basic HTTP client using aiohttp. Currently only GET is implemented. POST will be implemented if what I did so far isn't terrible.
Adds a lot more errors to `exceptions.py`. I'm not sure if these all should be used or if it should be abstracted more.

- `fixed`: Nothing :(

- `improvements`: None :(

 ### Check off the following

- [x] I have tested my changes with the current requirements
- [x] My Code follows the pep8 code style.


Sigmanificient (changes requested):
Can you use error a dictionary for the response status? Would reduce the amount the elif blocks

Sigmanificient (commented):
I think we could use `if  res.ok` for the first checks

Arthurdw (commented):
Improved structure and code quality + added extra http methods.

Lunarmagpie:
Currently error codes in the 5xx series that aren't 502 are unhandled so they'll silently fail. I think [this](https://github.com/Lunarmagpie/Pyscord/blob/main/pyscord/core/http.py#L110) should be an if/else block to get around that.
Also I think that the function should sleep for more time after each failed request. Afaik this is the more standard way to handle 5xx error codes (that's how discord.py did it at least).

pyscord/core/http.py
```diff
@@ -0,0 +1,145 @@
...
+
+            sender = methods.get(method)
+
+            if not sender:
+                raise RuntimeError("Invalid RequestMethod has been passed.")
+
+            async with sender(f"{self.endpoint}/{route}",
+                              headers=self.header) as res:
+
+                if res.status in [200, 201, 204]:
```
Sly-Little-Test, line 105: Maybe `res.status >= 200 and res.status < 300`?

Sly-Little-Test (changes requested):

pyscord/core/http.py
Sigmanificient: Response object does offer a `ok` property which is defined by the following:
```py
    @property
    def ok(self):
        """Returns True if :attr:`status_code` is less than 400, False if not.

        This attribute checks if the status code of the response is between
        400 and 600 to see if there was a client error or a server error. If
        the status code is between 200 and 400, this will return True. This
        is **not** a check to see if the response code is ``200 OK``.
        """
        try:
            self.raise_for_status()
        except HTTPError:
            return False
        return True
```

Which seems to be ok for what we doing since the status code can be grouped as:
- `100–199` Informational responses
- `200–299` Successful responses
- `300–399` Redirects
- `400–499` Client errors
- `500–599` Server errors
(*more @:* https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)

Lunarmagpie: Did this

pyscord/core/http.py
```diff
@@ -21,16 +21,35 @@
...
+from aiohttp.client import _RequestContextManager
+from aiohttp.typedefs import StrOrURL
 
-from ..exceptions import NotFoundError, BadRequestError, NotModifiedError, UnauthorizedError, ForbiddenError, MethodNotAllowedError, RateLimitError, ServerError
+from ..exceptions import NotFoundError, BadRequestError, NotModifiedError, \
+    UnauthorizedError, ForbiddenError, MethodNotAllowedError, RateLimitError, \
+    ServerError, HTTPError
+
+
+class RequestMethod(Enum):
```
Sigmanificient, line 37: I think `auto()` could be used for the id, right?

```py
from enum import Enum, auto

...

class RequestMethod(Enum):
    GET = auto()
    POST = auto()
    DELETE = auto()
    PUT = auto()
    OPTIONS = auto()
```

pyscord/core/http.py
```diff
@@ -39,48 +58,64 @@ def __init__(self, token: str, version: int = 9, max_retries: int = 4) -> None:
...
+            404: NotFoundError(),
+            405: MethodNotAllowedError(),
+            439: RateLimitError()
+        }
+
+    async def __send(self, method: RequestMethod, route: str, *,
+                     __ttl: int = None):
+        # TODO: Fix docs
+        # TODO: Implement logging
+        __ttl = self.max_ttl if __ttl is None else __ttl
```
Sigmanificient, line 83: I guess this could by simplified as
```py
__ttl = __ttl or self.max_ttl
```

pyscord/core/http.py
Sigmanificient, line 37: I think `auto()` could be used for the id, right?

```py
from enum import Enum, auto

...

class RequestMethod(Enum):
    GET = auto()
    POST = auto()
    DELETE = auto()
    PUT = auto()
    OPTIONS = auto()
```

pyscord/core/http.py
Sly-Little-Test: > 
> 
> I guess this could by simplified as
> 
> ```python
> __ttl = __ttl or self.max_ttl
> ```

Right.

PR #68, opened 2021-09-17
✨Guild Object with Client.get_guild

Endercheif:
### Changes

-   `adds`: 
    - `Client.get_guild(id)` to get a guild object
    - `Guild.from_id`
    - `ChannelType.from_id`
    - Extra Channel Types (Adding functionality is a TODO)
    - `_channel_type_map` to use the correct class
    - `Guild.nsfw` (missing from discord docs but in the v9 api)

Code has been tested with a bot.

 ### Check off the following

### DIFF
--- a/pincer/middleware/__init__.py
+++ b/pincer/middleware/__init__.py
@@ -22,6 +22,7 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 # SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import logging
 from glob import glob
 from importlib import import_module
 from os import chdir
@@ -30,6 +31,8 @@ from typing import Dict
 
 from pincer.exceptions import NoExportMethod
 from pincer.utils import Coro
+
+_log = logging.getLogger(__package__)
 
 
 def get_middleware() -> Dict[str, Coro]:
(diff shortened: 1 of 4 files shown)


pincer/middleware/__init__.py
```diff
@@ -48,6 +48,7 @@ def get_middleware() -> Dict[str, Coro]:
                 "export"
             )()
         except AttributeError:
+            continue # TODO: Fix this. Always raises error because some modules are empty
```
Sigmanificient, line 51: Can you add a debug log to let the users know what middlewares are missing for now? thx

pincer/objects/channel.py
```diff
@@ -34,6 +33,12 @@
 from .overwrite import Overwrite
 from ..utils import APIObject, APINullable, MISSING, Snowflake, Timestamp
 
+if TYPE_CHECKING:
+    from pincer import Client
```
Sigmanificient, line 37: doesn't `from . import Client` work here? i've seen you used `pincer` instead of `.` at multiples places

pincer/objects/channel.py
```diff
@@ -176,12 +188,40 @@ class Channel(APIObject):
     user_limit: APINullable[int] = MISSING
     video_quality_mode: APINullable[int] = MISSING
 
+    @classmethod
+    async def from_id(cls, client: Client, id: int) -> Channel:
+        data = (await client.http.get(f"/guilds/{id}")) or {}
+        data.update({"_client": client, "_http": client.http, "type": ChannelType(data.pop("type"))})
+        channel_cls = _channel_type_map.get(data["type"], Channel)
```
Sigmanificient, line 195: What do you think as something like:
```py
dataclass(type(_channel_type_map.get(data["type"], Channel), (Channel,), {})).from_dict(data)
```

pincer/objects/channel.py
Endercheif: I'm assuming `from . import Client` will work unless it's out of TYPE_CHECKING but I have not tested that yet.

PR #70, opened 2021-09-19
Yieldable commands

Endercheif:
### Changes

-   `adds`: 
     - Ability to have commands with `yield` (sends a webhook to respond). [see API docs for implementation](https://discord.com/developers/docs/interactions/receiving-and-responding#followup-messages)
     - When creating the command, it checks if it is coroutine or asyncgenerator
     - API Exception includes a JSON property to see the response when using try-except. (this is taken advantage of in this pr)
     - A separate function to convert a message to a `Message` object

 ### Check off the following

### DIFF
--- a/pincer/middleware/interaction_create.py
+++ b/pincer/middleware/interaction_create.py
@@ -21,11 +21,19 @@
 # CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 # SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+from typing import Union
+from asyncio import sleep
+from inspect import isasyncgenfunction
+import logging
+
 from pincer.commands import ChatCommandHandler
 from pincer.core.dispatch import GatewayDispatch
 from pincer.objects import Interaction, Embed, Message, InteractionFlags
+from pincer.exceptions import RateLimitError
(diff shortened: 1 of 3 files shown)


pincer/middleware/interaction_create.py
```diff
@@ -56,23 +64,59 @@ async def interaction_create_middleware(self, payload: GatewayDispatch):
...
+                        )
+
+                else:
+                    started = True
+
+                    await self.http.post(
+                        f"interactions/{interaction.id}/{interaction.token}/callback",
+                        msg.to_dict()
+                    )
+                await sleep(0.3)
```
Sigmanificient, line 96: why `await sleep(.3)`?

pincer/middleware/interaction_create.py
```diff
@@ -56,23 +64,59 @@ async def interaction_create_middleware(self, payload: GatewayDispatch):
...
+
+            await self.http.post(
+                f"interactions/{interaction.id}/{interaction.token}/callback",
+                message.to_dict()
+            )
 
     return "on_interaction_create", [interaction]
 
 
+def convert_message(self, message: Union[Embed, Message, str]) -> Message:
```
Sigmanificient, line 109: Extraction this to a function is a great idea. Tho, when we'll have more and more thing to convert to message (button, list, ...) we might want to use a dictionary and bind type key to convert function or something like?

pincer/middleware/interaction_create.py
Endercheif: The delay is needed otherwise there will be a RateLimitError. `0.3` is the "sweetspot" as `0.25` does not always help and `0.5` is too slow.

Issue #91, opened 2021-09-30
Argument descriptions


Arthurdw:
Implement argument descriptions.

eg.
```py
@command()
async def foo(argument: (type, "Description here")):
    ...
```

Issue #84, opened 2021-09-30
Make command execution future based


Arthurdw:
Title is pretty self explanatory. Making the lib a partly async and thread based model should greatly increase performance.

Arthurdw:
Note: Because of the way asyncio works I won't be implementing threading. But I'll do smth else to simulate the requested behavior.

PR #92, opened 2021-10-01
Improved the way events are executed + implemented command argument descriptions.

Arthurdw:
## Implementations

Implements #84 and #91.

### Usage 

```py
@command()
async def foo(bar: (str, "A sample command!")):
    ...
```

First tuple argument: type
Second tuple argument: description

Description defaults to `Description not set`

### DIFF
--- a/pincer/client.py
+++ b/pincer/client.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import logging
-from asyncio import iscoroutinefunction, run
+from asyncio import iscoroutinefunction, run, ensure_future
 from inspect import isasyncgenfunction
 from typing import Optional, Any, Union, Dict, Tuple, List
 
@@ -349,11 +349,11 @@ class Client(Dispatcher):
             The named arguments for the event.
         """
         if call := self.get_event_coro(name):
-            await self.execute_event(call, error, *args, **kwargs)
+            self.execute_event(call, error, *args, **kwargs)
         else:
             raise error
 
-    async def execute_event(self, call: Coro, *args, **kwargs):
+    def execute_event(self, call: Coro, *args, **kwargs):
         """
         Invokes an event.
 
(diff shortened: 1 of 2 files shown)


pincer/client.py
```diff
@@ -366,11 +366,12 @@ async def execute_event(self, call: Coro, *args, **kwargs):
...
             The named arguments for the event.
         """
+        def execute(*_args, **_kwargs):
+            ensure_future(call(*_args, **_kwargs))
 
-        if should_pass_cls(call):
-            await call(self, *args, **kwargs)
-        else:
-            await call(*args, **kwargs)
+        execute(self, *args, **kwargs) \
```
Sigmanificient, line 372: Can you change the style of this bloc?

I feel like the actual way it is written could be improved to not have backslashes.

I tend to use ternary only when using a `return` or assigning to a variable, cause i find the classic if/else more readable otherwise, what do you think?

Sigmanificient, line 372: Can you change the style of this bloc?

I feel like the actual way it is written could be improved to not have backslashes.

I tend to use ternary only when using a `return` or assigning to a variable, cause i find the classic if/else more readable otherwise, what do you think?

Arthurdw: So you'd like this change to be reverted?

Sigmanificient: in my opinion, a `if`/`else` seems better for that block

Arthurdw: Actually, I'll replace it with this. As it makes the code and its intent more clear.

```py
if should_pass_cls(call):
    args = [self, *args]

execute(*args, **kwargs)
```

Sigmanificient (approved):

pincer/client.py
```diff
@@ -371,7 +371,7 @@ def execute(*_args, **_kwargs):
             ensure_future(call(*_args, **_kwargs))
 
         if should_pass_cls(call):
-            args = [self, *args]
+            args = (self, *args)
```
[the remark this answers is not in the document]
Arthurdw: Yesh!

PR #94, opened 2021-10-01
🔖 0.7.1 Support for argument descriptions

Sigmanificient:
This version implement the support for argument descriptions.

## Examples

> before `0.7.1` *still suppported*
```py
@command(description="Add two numbers!")
async def add(
        first: int,
        second: int
):
    return f"The addition of `{first}` and `{second}` is `{first + second}`"
```

> argument descriptions
```py
@command(description="Add two numbers!")
async def add(
        first: (int, "The first number"),
        second: (int, "The second number")
):
    return f"The addition of `{first}` and `{second}` is `{first + second}`"
```

## Result within discord
image

### DIFF
--- a/.github/workflows/pypi_upload.yaml
+++ b/.github/workflows/pypi_upload.yaml
@@ -19,6 +19,9 @@ jobs:
           python-version: '3.x'
           architecture: 'x64'
 
+      - name: install requirements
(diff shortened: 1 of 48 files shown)


Arthurdw (approved):
Everything seems fine now.

PR #95, opened 2021-10-02
Implements command choices

Arthurdw:
## Implements

Implements command argument choices.

## Example

You can add choices to arguments through the `pincer.Choices` *(which is an alias for `typing.Literal`)*.
Choice values must consist of the same type, and there can only be a maximum of 25 choices.

```py
@command()
async def choose(self, choice: Choices["Test", "Sample lol", "Geheh"]):
    return f"You chose: `{choice}`"
```

image

### Example with custom name/value

You can manually override the default name (which is set to the value) of a choice.
This is done through a tuple, where the first value is the choice name and the second one is the value. These can be combined with default choices. *(so no tuple, as seen in the example)*

```py
@command()
async def choose(
        self,
        choice: Choices[
            ("This will be the test value", "Test"),
            "Sample lol",
            "Geheh"
        ]
):
    return f"You chose: `{choice}`"
 ```
 
image

*(So when `This will be the test value` is selected the bot will return `You chose: 'Test'`)*

### DIFF
--- a/pincer/__init__.py
+++ b/pincer/__init__.py
@@ -20,6 +20,7 @@ __author__ = "Sigmanificient, Arthurdw"
 __email__ = "contact@pincer.org"
 __license__ = "MIT"
 
+from pincer.utils import Choices
 
 ReleaseType = Optional[Literal["alpha", "beta", "candidate", "final", "dev"]]
 
@@ -46,5 +47,5 @@ class VersionInfo(NamedTuple):
 __version__ = VersionInfo(0, 7, 1)
 __all__ = (
     "__author__", "__email__", "__package__", "__title__",  "__version__",
-    "Bot", "Client", "command", "Intents"
+    "Bot", "Client", "command", "Intents", "Choices"
 )
--- a/pincer/utils/types.py
+++ b/pincer/utils/types.py
@@ -1,7 +1,7 @@
 # Copyright Pincer 2021-Present
 # Full MIT License can be found in `LICENSE` at the project root.
 
-from typing import TypeVar, Callable, Coroutine, Any, Union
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: pincer/__init__.py [hunk 1]
```diff
@@ -1,50 +1,51 @@
...
-            )
-        )
-
-
-__version__ = VersionInfo(0, 7, 1)
-__all__ = (
-    "__author__", "__email__", "__package__", "__title__",  "__version__",
-    "Bot", "Client", "command", "Intents"
-)
+"""
```
Sigmanificient, line 1: what happening with that file?

pincer/utils/types.py
```diff
@@ -1,20 +1,24 @@
...
+
+# Represents a value which is optionally returned from the API
+APINullable = Union[T, MissingType]
+
+# Represents a coroutine.
+Coro = TypeVar("Coro", bound=Callable[..., Coroutine[Any, Any, Any]])
+
+Choices = Literal
+
+choice_value_types = (str, int, float)
```
Sigmanificient, line 24: Shouldn't put directly `ChoiceValues = Union[str, int, float` ?

Sigmanificient, line 24: Shouldn't put directly `ChoiceValues = Union[str, int, float]` ?

pincer/__init__.py
Arthurdw: Just imported `Choices`

pincer/utils/types.py
Arthurdw: No, makes type checking easier.

Sigmanificient (approved):

PR #97, opened 2021-10-02
🔖 `0.7.2` Adding command choices

Sigmanificient:
This version add the support for command choice.

```py
    @command()
    async def choose(
            self,
            choice: Choices[
                ("This will be the test value", "Test"),
                "Sample lol",
                "Geheh"
            ]
    ):
        return f"You chose: `{choice}`"
```

image

image


Arthurdw (commented):
~ Good for prod

Issue #87, opened 2021-09-30
✨ Support for `tasks.loop`


Sigmanificient:
## Task
Make a decorator that transform a asynchronous function to a task object. 

```py
@task.loop(minute=30)
async def my_task(self):
    ...
```

When calling `my_task.start()` this will register the task to repeat the given amount of time.

mwath:
It sounds very useful!
I would love to contribute to this project starting with this issue if you don't mind.

Arthurdw:
> It sounds very useful! I would love to contribute to this project starting with this issue if you don't mind.

We don't mind, all contributions are welcome! 👐

PR #98, opened 2021-10-02
✨ Implements a task system

mwath:
# Implements
Implements a task system for #87 

### Changes
-   `adds`: `pincer.utils.tasks`

# Example
You can create a task that runs every 30 minutes, or any other given time.

```Python
from pincer import Client
from pincer.utils import TaskScheduler

client = Client("token")
task = TaskScheduler(client)

@task.loop(minutes=30)
async def my_task(self):
    ...

my_task.start()
client.run()
```

Furthermore, any of the following arguments can be specified to define the delay between each execution:
- `days`
- `weeks`
- `hours`
- `minutes`
- `seconds`
- `microseconds`
- `milliseconds`

 # Check off the following


pincer/exceptions.py
```diff
@@ -133,6 +133,22 @@ def from_desc(cls, _type: str, max_size: int, cur_size: int):
         )
 
 
+class TaskAlreadyRunning(PincerError):
```
Sigmanificient, line 136: Can you make a `TaskError` parent class that will inherit from `PincerError` and will be inherited by all the related error you added? Could be useful. Maybe could even have a attribute that link the the task that cause the error?

pincer/utils/tasks.py
```diff
@@ -0,0 +1,158 @@
...
+        """Check if the task has been cancelled or not."""
+        return self.running and self._handle.cancelled()
+
+    @property
+    def running(self):
+        """Check if the task is running."""
+        return self._handle is not None
+
+    def start(self):
+        """Register the task in the TaskScheduler and start the execution of the task."""
```
Sigmanificient, line 55: Make sure to respect the `pep8 & E501` for the docstring ^^
Seems hard but this is nothing crazy, just wrap the docstring and make the comments to have less than 72 characters, and have to code < 80 characters per line too.

If you are using pycharm, you can install the `.editorconfig` extension, which will allow you to load the config of the `.editorconfig` file with jb specific rules, that will show the 2 lines at 72 & 80 characters.

Sigmanificient (changes requested):
Hi, everything seems to look alright within the code, with only very minor tweaks. 
I will take the time to test the tasks myself too and wait for Arthur review.

Sigmanificient:
Oh, btw, if you can use [Gitmoji](https://gitmoji.dev/) for the next commits, would be really appreciated.
This is just a way to organize the commit a by categories with some GitHub emojis.

If you don't want tho, there is no problem, but I will have to squash the commit, which i would prefer avoid tbh, as would reduce the number of commit in your activity.

pincer/utils/tasks.py
mwath: Oops I'm sorry, I should have been more careful about it.
Thank you for the tip but I'm using vscode ^^ I'll add a ruler so I can spot the limit right away.

pincer/exceptions.py
mwath: That's a good idea actually 🤔 
I think I'll add it inside the `TaskError`class. However half errors are raised before the task object is created, so it might be None.

Sigmanificient: maybe just the task function and decorator argument then? feel free to add what seems the best for you

pincer/utils/tasks.py
Sigmanificient: no problem

pincer/utils/tasks.py
```diff
@@ -0,0 +1,158 @@
...
+        self._loop = asyncio.get_event_loop()
+
+    def loop(
+        self,
+        days=0,
+        weeks=0,
+        hours=0,
+        minutes=0,
+        seconds=0,
+        microseconds=0,
```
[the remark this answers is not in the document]
mwath: I don't think so either, but I'm using `timedelta` under the hood which support both. I thought it would be better to include it anyway 🤷‍♂️

pincer/exceptions.py
mwath: Both the error and the traceback are straight-forward when a `TaskIsNotCoroutine` or a `TaskInvalidDelay` error is raised. Including more information is not really relevant.

pincer/utils/tasks.py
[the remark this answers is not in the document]
Arthurdw: Yeah, having them doesn't hurt :)

Arthurdw (approved):
Thank you for your amazing contributions! ♥

pincer/exceptions.py
```diff
@@ -133,6 +133,32 @@ def from_desc(cls, _type: str, max_size: int, cur_size: int):
         )
 
 
+class TaskError(PincerError):
```
Sigmanificient, line 136: Great!

Sigmanificient (approved):

PR #109, opened 2021-10-03
📝 Add/update a few docstrings

Sriram-bb63:
### Changes

-   `adds`: docstrings for classes and methods

 ### Check off the following


pincer/utils/directory.py
```diff
@@ -6,7 +6,7 @@
 
 @contextmanager
 def chdir(path) -> Generator[str, Any, None]:
-    """Change the cwd to the given path as a context manager."""
+    """Change the cwd to the given path as a context manager"""
```
Sigmanificient, line 9: Can you add back the dot please? Also put it on the other ones ^^

> Quoted from [**PEP 257**](https://www.python.org/dev/peps/pep-0257/)

- *The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".*

Sigmanificient (changes requested):

pincer/utils/directory.py
Sriram-bb63: Yes, sure. TIL end docstrings with a dot.

Sigmanificient (approved):

PR #110, opened 2021-10-04
Added support for File attachments

Lunarmagpie:
### Changes

-   `adds`: Files.
There is now support for any bytes. Pillow JPG and PNG images and files have helper functions. Other formats can be sent by constructing a file object manually.

Example
Pillow image
```python
class Bot(Client):
    
    @command(description="Test method")
    async def test(self):

        img = Image.new("RGB",(500,500),(20,20,20))

       # Method one if you want to specify an image format. The file extension is used as the format.
        return Message(attachments=[
            File.from_image(
                img,
                "image.png"
            )
        ])

      #Method 2 & 3, can only do PNGs.
      return img
      return Message(attachments=[img])
```
Sending a file

```python
class Bot(Client):
    
    @command(description="Test method")
    async def test(self):

       # Method one if you want to specify an image format. The file extension is used as the format.
        return Message(attachments=[
            File.from_file(
                 "test/something.txt"
                  #Filename can be specified if wanted
                 filename="somethingelse.txt"
            )
        ])

      # Can also be done like a png
      return Message(attachments=["test/something.txt"])
      
      # but this returns a message. Someone is gonna be baited by this. Probably me :(.
      return "something.txt"
      
```

-   `fixed`: Nothing :(
-   `improvements`: the FIle object is used when a message is sent to discord. Attachment is used when it is received. This is because File uses bytes to store the data but Attachment wants a URL.

 ### Check off the following


pincer/objects/message/file.py
```diff
@@ -0,0 +1,69 @@
...
+    def from_image(
+        cls,
+        img: Image,
+        filename: str,
+        format: Optional[str] = None
+    ) -> File:
+        """
+        Creates a file object from a PIL image
+        Supports PNG and JPEG
+
```
Sigmanificient, line 52: documentation is missing params

pincer/objects/message/message.py
```diff
@@ -1,79 +1,129 @@
...
+    flags: Optional[InteractionFlags] = None
+    type: Optional[CallbackType] = None
+
+    def __post_init__(self):
+
+        attch = []
+
+        for count,value in enumerate(self.attachments):
+            if isinstance(value,File):
+                attch += [value]
```
Sigmanificient, line 68: `+=` could be replace with `.extend` and `.append`

pincer/objects/message/message.py
```diff
@@ -1,79 +1,129 @@
...
+        }
+
+    def serialize(self) -> Tuple[str,Union[Payload,Dict]]:
+        """
+        Creates the data that the discord API wants for the message object
+
+        :return: (content_type, data)
+        """
+
+        if self.attachments:
```
Sigmanificient, line 118: invert the condition to put the else `return` at top in order to flatten the code

```py
if ...:
   # stuff
   ...
   # more stuff
   ...
   # even more stuff
   ...
else:
   return ...

```

can be simplified as:
```py
if not ...:
    return ...

# stuff
...
# more stuff
...
# even more stuff
...
```

Why do this? I make the most important part less indented and put specific cases at the top, so the can be found and read faster

pincer/objects/message/message.py
```diff
@@ -1,79 +1,129 @@
...
+            form = FormData()
+            form.add_field("payload_json",json.dumps(self.to_dict()))
```
Sigmanificient, line 120: i believe `form` could have chaining, and maybe a `.add_fields` method. This would allow us to write someone as 
```py
return (
    FormData()
        .add_field(...)
        .add_fields(self.attachments, lambda file: ("file",file.content, file.filename))
).headers["Content-Type"], payload

```
here.

Sigmanificient (changes requested):

Endercheif:
also change instances of `x,y` to `x, y` (notice the extra space)

Sigmanificient:
> also change instances of `x,y` to `x, y` (notice the extra space)

Lunarmagpie didn't checked the `pep8` style so i didn't look at the style for now

requirements.txt
```diff
@@ -1,2 +1,3 @@
-websockets
-aiohttp
\ No newline at end of file
+aiohttp
```
Sigmanificient, line 1: Can you put specific more versions?

pincer/middleware/interaction_create.py
```diff
@@ -20,6 +22,8 @@ def convert_message(self, message: Union[Embed, Message, str]) -> Message:
     """Converts a message to a Message object"""
     if isinstance(message, Embed):
         message = Message(embeds=[message])
+    elif isinstance(message, File) or isinstance(message, Image):
```
Sigmanificient, line 25: You can merge the instance call here
```diff
-     elif isinstance(message, File) or isinstance(message, Image):
+    elif isinstance(message, (File, Image)):
```

pincer/objects/message/message.py
```diff
@@ -1,79 +1,128 @@
...
+            "users": list(map(get_str_id, self.users)),
+            "replied_user": self.reply
+        }
+
+
+@dataclass
+class Message:
+    # TODO: Write docs
+    content: str = ''
+    attachments: Optional[File] = None
```
Sigmanificient, line 54: Isn't `attachements` type `Optional[List[File]]` ?
Maybe even `ApiNullable[List[File]]`

Sigmanificient (changes requested):

pincer/middleware/interaction_create.py
Lunarmagpie: lol, just saw this a second ago in the aiohttp source. Was just about to change it.

pincer/objects/message/file.py
```diff
@@ -0,0 +1,69 @@
...
+            content=file,
+            filename=filename or os.path.basename(filepath)
+        )
+
+    @classmethod
+    def from_image(
+            cls,
+            img: Image,
+            filename: str,
+            format: Optional[str] = None
```
Sigmanificient, line 47: can you use an other name than `format` to avoid shadow scope?

Sigmanificient (changes requested):

pincer/objects/message/message.py
Lunarmagpie: Yeah, I will set it to Optional[List[File]] = None since thats the format for the other params for some reason.

pincer/objects/message/message.py
Lunarmagpie: It doesn't look like it has chaining.
I can replace the for loop with add_fields() but Im not sure which is more readable.

```python
        form.add_fields(
            *(
                (file.filename, file.content)
                for file in self.attachments
            )
        )
```
```python
        for file in self.attachments:
            form.add_field("file", file.content, filename=file.filename)
```

requirements.txt
Lunarmagpie: Yeah, I added the version number for all of the libraries we use.

pincer/objects/message/message.py
Sigmanificient: I mean that why i said `could` haha, yes unfortunately doesn't hav, but may be not to hard to add.
From your 2 proposition, the 2nd one sound more readable

pincer/objects/message/file.py
```diff
@@ -0,0 +1,91 @@
...
+        :param filepath:
+            The path to the file you want to send. Must be string. The file's
+            name in the file path is used as the name when uploaded discord by
+            default.
+
+        :param filename:
+            The name of the file. Will override the default name.
+        """
+
+        file = open(filepath, "rb").read()
```
Sigmanificient, line 44: unclosed file? why not using 

```py
with open(filepath, "rb") as f:
    file = f.read()
```

pincer/objects/message/message.py
```diff
@@ -1,79 +1,134 @@
...
+                ))
+            elif isinstance(value, str):
+                attch.append(File.from_file(value))
+            else:
+                raise ValueError(f"Attachment {count} is invalid type.")
+
+        self.attachments = attch
+
+    def to_dict(self):
+        if len(self.content) < 1 and not self.embeds and not self.attachments:
```
Sigmanificient, line 85: may want to extract this to a `__bool__` method?
such as `if not self:`

pincer/objects/message/message.py
Lunarmagpie: Im a little confused about what you want. I didn't write the class so Im at the mercy of the aiohttp gods.

pincer/objects/message/message.py
Lunarmagpie: What if I make a method like `self.is_empty()`? Just so its more explicit.

Sigmanificient (approved):

pincer/objects/message/message.py
Sigmanificient: A property them, could be

pincer/objects/message/file.py
```diff
@@ -0,0 +1,94 @@
...
+        with open(filepath, "rb") as data:
+            file = data.read()
+
+        return cls(
+            content=file,
+            filename=filename or os.path.basename(filepath)
+        )
+
+    @classmethod
+    def from_image(
```
Arthurdw, line 53: Could this be renamed to `from_pillow_image`?

Lunarmagpie: Sure. I can change this in a few hours.

pincer/objects/message/message.py
Lunarmagpie: I think I will change it to `isempty()` actually because this keeps in line with the python methods listed here https://stackoverflow.com/questions/47996388/python-boolean-methods-naming-convention/47996592

Sigmanificient: I usually tend to make boolean method properties are there are generally low cost
About the `is` + `<word>` i always used `_` but now that i think about it i don't know which is better

pincer/objects/message/file.py
Lunarmagpie: Done

pincer/objects/message/file.py
```diff
@@ -82,11 +83,17 @@ def from_image(
             if image_format == "jpg":
                 image_format = "jpeg"
 
+<<<<<<< HEAD
```
Sigmanificient: you have a merge conflict right here

Sigmanificient: you have a merge conflict right here

pincer/objects/message/message.py
```diff
@@ -81,7 +99,7 @@ def __post_init__(self):
 
         self.attachments = attch
 
-    def is_empty(self) -> bool:
+    def isempty(self) -> bool:
```
Sigmanificient, line 102: can you make it a property?

Sigmanificient (changes requested):

pincer/objects/message/message.py
Lunarmagpie: sure. Ill do `empty`.

Sigmanificient (approved):

Sigmanificient (approved):

PR #114, opened 2021-10-04
✨ Created a COG system

Arthurdw:
Created a cog system where the user can load cogs & created an example.

### Basic example:

`run.py`
```py
>>> from pincer import Client
>>>
>>> class BotClient(Client):
```
`cogs/say.py`
```py
>>>
>>> class SayCommand:
...     @command()
...         return message
>>>
>>> setup = SayCommand
```

#### Note on setup method

```py
def setup(client):
    return MyCog(client.bot.name)
 ```

### CODE REVIEW

File: pincer/client.py [hunk 1]
```diff
@@ -14,11 +15,13 @@
 from .core.dispatch import GatewayDispatch
 from .core.gateway import Dispatcher
 from .core.http import HTTPClient
-from .exceptions import InvalidEventName
+from .exceptions import InvalidEventName, TooManySetupArguments, \
```
Sigmanificient, line 18: Just a parenthesis would be nice here :D

pincer/utils/signature.py
```diff
@@ -0,0 +1,38 @@
...
+        func = getattr(func, "__init__")
+
+        if func is object.__init__:
+            return [], []
+
+    sig = signature(func).parameters
+    params = list(sig)
+
+    if should_pass_cls(func):
+        del params[0]
```
Sigmanificient, line 26: `del params[0]` vs `params.pop(0)` ?

Sigmanificient (changes requested):

Sigmanificient (approved):

Sigmanificient (approved):

PR #116, opened 2021-10-04
🔖 0.9.0 Cogs & Channel Edition

Sigmanificient:
The `0.9.0` implements the cogs system!

*very original content i definitively wrote:*
### Basic example:

`run.py`

```py
>>> from pincer import Client
>>>
>>> class BotClient(Client):
```

`cogs/say.py`

```py
>>>
>>> class SayCommand:
...     @command()
...         return message
>>>
>>> setup = SayCommand
```

**Note on setup method**

```py
def setup(client):
    return MyCog(client.bot.name)
```

## Channel Edition
*yes i even stole text from my previous issues!*

Implemented a `channel.edit` method. 
In this method you can pass named arguments for the channel attribute(s) that have to change.

### Example:
```py
@command()
async def update_channel():
    channel = await Channel.from_id(...)
    
    # the method to add
    await channel.edit(name="foo")

```


Arthurdw (approved):
~Ready for merge

Sigmanificient:
> 
> 
> ~Ready for merge

Oh NO NO NO ... it happened... again xD

PR #120, opened 2021-10-04
🔖 0.9.0 Cogs & Channel Edition [FULL]

Sigmanificient:
The `0.9.0` implements the cogs system!

*very original content i definitively wrote:*
### Basic example:

`run.py`

```py
>>> from pincer import Client
>>>
>>> class BotClient(Client):
```

`cogs/say.py`

```py
>>>
>>> class SayCommand:
...     @command()
...         return message
>>>
>>> setup = SayCommand
```

**Note on setup method**

```py
def setup(client):
    return MyCog(client.bot.name)
```

## Channel Edition
*yes i even stole text from my previous issues!*

Implemented a `channel.edit` method. 
In this method you can pass named arguments for the channel attribute(s) that have to change.

### Example:
```py
@command()
async def update_channel():
    channel = await Channel.from_id(...)
    
    # the method to add
    await channel.edit(name="foo")

```


pincer/client.py
```diff
@@ -5,20 +5,26 @@
 
 import logging
 from asyncio import iscoroutinefunction, run, ensure_future
+from importlib import import_module
 from inspect import isasyncgenfunction
 from typing import Optional, Any, Union, Dict, Tuple, List
 
+from pincer.objects import AppCommand
```
Arthurdw, line 12: Change this to relative import

Arthurdw: If its even used :?

Arthurdw (changes requested):
Relative import if required

pincer/client.py
Sigmanificient: used as a typehint at line 371 `to_remove: List[AppCommand] = []`

pincer/exceptions.py
```diff
@@ -24,8 +24,49 @@ def __init__(self, specific: str):
 
 class NoExportMethod(PincerError):
     """
-    Exception which gets raised when an `export` method is expected but not
-    found in a module.
+    Exception which gets raised when an `export` method is expected but
```
Endercheif, line 27: can this be changed from \`export\` to \`\`export\`\` so that it works with the docs

pincer/exceptions.py
```diff
@@ -24,8 +24,49 @@ def __init__(self, specific: str):
...
+class CogAlreadyExists(CogError):
+    """
+    Exception which gets raised when a cog is already loaded, but is
+    trying to be reloaded!
+    """
+
+
+class NoValidSetupMethod(CogError):
+    """
+    Exception which gets raised when an `setup` function is expected but
```
Endercheif, line 54: \`\`setup\`\`

pincer/middleware/ready.py
```diff
@@ -23,7 +23,7 @@ async def on_ready_middleware(self, payload: GatewayDispatch):
         The data received from the ready event.
     """
     self.bot = User.from_dict(
-        {"_client": self, "_http": self.http,  **payload.data.get("user")}
+        {"_client": self, "_http": self.http, **payload.data.get("user")}
```
Endercheif, line 26: If `payload.data.get("user")` returns None, then this would cause an error.

Arthurdw: Resolved, now it will throw a `InvalidPayload`

pincer/exceptions.py
Arthurdw: Fixed

pincer/exceptions.py
Arthurdw: Fixed

pincer/client.py
Arthurdw: Resolved with `TYPE_CHECKING`

Arthurdw (approved):
~ Ready for merge this time

PR #124, opened 2021-10-05
82 implement middleware

krishnajalan:
### Changes

-   `adds`: Voice_state_update, Guild_create, message_update Middleware
-   `fixed`: added missing fields in Guild that were causing errors in Guild_create middleware
-   `improvements`: added missing fields in voicestate structure

 ### Check off the following

-   [ ] My Code follows the pep8 code style.

### DIFF
--- a/pincer/objects/guild/guild.py
+++ b/pincer/objects/guild/guild.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from enum import Enum, auto, IntEnum
-from typing import Optional, List, overload, TYPE_CHECKING
+from typing import Optional, List, Dict, overload, TYPE_CHECKING
 
 from .member import GuildMember
 from ..events.presence import PresenceUpdateEvent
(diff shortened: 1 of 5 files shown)


pincer/objects/guild/guild.py
```diff
@@ -414,6 +414,9 @@ class Guild(APIObject):
     afk_channel_id: Optional[Snowflake]
     afk_timeout: int
     application_id: Optional[Snowflake]
+    application_command_counts: Optional[Dict]
```
Arthurdw, line 417: @Sigmanificient I think we should use the `APINullable` here with a default of `MISSING` not?

Arthurdw:
~ Removed link as more middleware has to be implemented

pincer/objects/guild/guild.py
Sigmanificient: Indeed, i believe

krishnajalan: Updated the code as suggested.

Sigmanificient:
> 
> 
> ~ Removed link as more middleware has to be implemented

Will add back the link afterward then, as still a part of the issue

PR #140, opened 2021-10-10
Implemented more command types

zunda-arrow:
### Changes
Implemented String, Integer, Boolean, Number, User, Channel, and Role application command option types.
### Examples:
```python
@command(description="method 2")
# number will now be a float instead of string 
async def number(self, number: float):
    return Message("The number is %s" % number)

# A User object will now be returned
async def number(self, user: User):
    return Message("The user's name is %s" % user.username)
```

#### New Options
```
str - String
int - Integer
bool - Boolean
float - Number
pincer.User - User
pincer.Channel - Channel
pincer.Role - Role
```

### Notes
I think the if/else block in the `set_param()` function is a little bit of a mess. I didn't know what to do about it so please tell me ideas if you have them.

 ### Check off the following


Lunarmagpie:
I wasn't able to reopen because I force pushed to the main branch :( I guess I made a dead pull request.

PR #173, opened 2021-11-02
:children_crossing: Allow non-parenthesised `@command` decorator

beastmatser:
### Changes

This small change simply allows for non-parenthesised `@command` decorators

```py
@command  # () is optional
async def test():
   return Message("Succes!")
```

I accidentally committed some stuff to the wrong branch, so I closed #168 since I had to revert that commit.

 ### Check off the following

### DIFF
--- a/pincer/commands.py
+++ b/pincer/commands.py
@@ -8,6 +8,7 @@ import re
 from asyncio import iscoroutinefunction, gather
 from copy import deepcopy
 from inspect import Signature, isasyncgenfunction
+from functools import partial
 from typing import (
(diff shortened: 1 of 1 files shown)


Arthurdw (approved):

Sigmanificient (approved):
The conflicts are going to be a pain to solve

beastmatser:
I feel like it would be better to close this for now and maybe do this at a later point or not to it at all

beastmatser:
This will be added again in a new pull request, this easier than to fix the merge conflicts

PR #176, opened 2021-11-04
✨ Added Client.wait_for() and Client.loop_for()

Lunarmagpie:
### Changes

-   `adds`: ...
`Client.wait_for()` - Same functionality as discord.py.
`Client.loop_for()` - Returns async generator that iterates when an event happens. See guessing_game.py example.
Both these functions return `*args` from the respective middleware. I ignored **kwargs since it makes deconstructing the tuple it returns really annoying. I think we should change all the middleware to return a single object in the near future instead of a list for this reason. We don't even use more than one item in the list right now.
-   `improvements`: You can now type hint with MessageContext in slash commands and it will be ignored

 ### Check off the following


examples/guessing_game.py
```diff
@@ -0,0 +1,42 @@
...
+        await ctx.reply(f"Starting the guessing game! Picker a number between 0 and {biggest_number}.")
+        channel = await self.get_channel(ctx.channel_id)
+        number = random.randint(0, biggest_number)
+
+        try:
+            async for next_message, in self.loop_for('on_message'):
+                if next_message.author.bot:
+                    continue
+
+                try:
```
Sigmanificient, line 21: i think you should use `.isdigit()`  here to remove one `try/except` level

pincer/client.py
```diff
@@ -547,6 +551,63 @@ async def payload_event_handler(self, _, payload: GatewayDispatch):
         """
         await self.process_event("payload", payload)
 
+    async def wait_for(
+        self,
+        event_name: str,
+        check: Optional[Callable[[Any], bool]] = None,
+        timeout: Optional[float] = None
```
Sigmanificient, line 558: is timeout as a `float` necessary, like don't seems necessary to put a timeout of `.5`  instead of `1`...
float seems kinda overkill

pincer/core/http.py
```diff
@@ -306,7 +306,7 @@ async def options(self, route: str) -> Optional[Dict]:
     async def patch(
             self,
             route: str,
-            data: Dict,
+            data: Dict = None,
```
Sigmanificient, line 309: `Optional[Dict]`

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,211 @@
...
+    ----------
+    return_value : Optional[str]
+        Used to store the arguments from ``can_be_set`` so they can be
+        returned later.
+    """
+
+    def __init__(
+        self,
+        event_name: str,
+        check: Union[Callable[[Any], bool], None]
```
Sigmanificient, line 20: maybe extract the type to a typealiase like `Check` or something

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,211 @@
...
+class EventMgr:
+    """
+    Attributes
+    ----------
+    stack : List[_DiscordEvent]
+        The List of events that need to be processed.
+    """
+
+    def __init__(self):
+        self.stack = []
```
Sigmanificient, line 70: @Arthurdw Using custom stack class maybe? not sure

Sigmanificient (changes requested):

pincer/client.py
Lunarmagpie: It is since it's in seconds. Also, it just passes through to `asyncio.wait_for()` that uses floats.

pincer/utils/event_mgr.py
Lunarmagpie: The class uses methods from the list. (`list.remove()`). I will rename the variable so it makes more sense.

examples/guessing_game.py
```diff
@@ -0,0 +1,42 @@
...
+                    continue
+
+                if guessed_number > number:
+                    await channel.send("Number is too high!")
+                elif guessed_number < number:
+                    await channel.send("Number is too low!")
+                else:
+                    await next_message.react("🚀")
+                    await channel.send("Number is correct!")
+                    self.game_running = False
```
trag1c, line 34: `self.game_running` doesn't seem to have any use?

Lunarmagpie: fixed :pensive:

trag1c (approved):

Lunarmagpie:
I realized theres a huge issue with `Client.loop_for()`. It can miss a message if the function is yielding while a new message is received. It can avoided by using the filter parameter but I think this issue is big enough the entire function.

Lunarmagpie:
I think this is good to review

pincer/client.py
```diff
@@ -581,6 +586,63 @@ async def payload_event_handler(self, _, payload: GatewayDispatch):
...
+        """
+        Parameters
+        ----------
+        event_name : str
+            The type of event. It should start with `on_`. This is the same
+            name that is used for @Client.event.
+        check : Callable[[Any], bool], default=None
+            This function only returns a value if this return true.
+        timeout: Union[float, None]
+            Amount of seconds before timeout.
```
trag1c, line 604: Should replace the old typehints
```suggestion
        check : CheckFunction
            This function only returns a value if this return true.
        timeout: Optional[float]
            Amount of seconds before timeout.
```

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,287 @@
...
+    value. ``None`` is returned if all arguments are ``None``.
+    """
+    args_without_none = [n for n in args if n is not None]
+
+    if len(args_without_none) == 0:
+        return None
+
+    return min(
+        args_without_none
+    )
```
trag1c, line 62: wat doink
```suggestion
    return min(args_without_none)
```

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,287 @@
...
+        if self.matches_event(event_name, *args):
+            self.events.append(args)
+            self.wait.set()
+
+    async def get_next(self):
+        """
+        Returns the next item if the queue. If there are no items in the queue,
+        it will return the next event that happens.
+        """
+        if len(self.events) == 0:
```
trag1c, line 153:
```suggestion
        if not self.events:
```

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,287 @@
...
+class EventMgr:
+    """
+    Attributes
+    ----------
+    event_list : List[_DiscordEvent]
+        The List of events that need to be processed.
+    """
+
+    def __init__(self):
+        self.event_list: _Processable = []
```
trag1c, line 173: Is that definitely the right typehint?

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,287 @@
...
+        Returns
+        ------
+        Any
+            What the Discord API returns for this event.
+        """
+
+        event = _Event(
+            event_name=event_name,
+            check=check
+        )
```
trag1c, line 213: Don't think it's the best idea to use keyword arguments if they're not required here
```suggestion
        event = _Event(event_name, check)
```

pincer/utils/event_mgr.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+        self.event_list.remove(event)
+        return event.return_value
+
+    async def loop_for(
+        self,
+        event_name: str,
+        check: Union[Callable[[Any], bool], None],
+        iteration_timeout: Union[float, None],
+        loop_timeout: Union[float, None],
```
trag1c, line 230: Update typehints

trag1c (changes requested):

pincer/utils/event_mgr.py
Lunarmagpie: Im sure this is the right one. `event_list` stores the `_LoopMgr` class and `_Event` class since the client can run the exact same functions on both of them.

trag1c (approved):

pincer/utils/event_mgr.py
trag1c: What I meant was why is this typed as `_Processable` and then set to a list. Did you mean `list[_Processable]`?

pincer/utils/event_mgr.py
Lunarmagpie: oh im still stupid

Sigmanificient:
raise:
```py
Traceback (most recent call last):
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 570, in event_handler
    await self.process_event(payload.event_name.lower(), payload)
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 553, in process_event
    await self.execute_error(e)
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 527, in execute_error
    raise error
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 545, in process_event
    key, args, kwargs = await self.handle_middleware(payload, name)
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 479, in handle_middleware
    extractable = await ware(self, payload, *args, **kwargs)
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/client.py", line 115, in wrapper
    return await (
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/middleware/interaction_create.py", line 159, in interaction_create_middleware
    raise e
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/middleware/interaction_create.py", line 140, in interaction_create_middleware
    await interaction_handler(self, interaction, context,
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/middleware/interaction_create.py", line 101, in interaction_handler
    await interaction_response_handler(
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/middleware/interaction_create.py", line 65, in interaction_response_handler
    message = await command(**kwargs)
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/examples/guessing_game.py", line 21, in guess
    async for next_message, in self.loop_for('on_message'):
  File "/home/sigmanificient/PycharmProjects/Lunar-Pincer/pincer/utils/event_mgr.py", line 281, in loop_for
    loop_timeout -= loop.time() - start_time
TypeError: unsupported operand type(s) for -=: 'NoneType' and 'float'
```

PR #191, opened 2021-11-11
Allow cogs to be loaded by callable instead of filename

mCodingLLC:
Not sure if this is an appropriate way to solve the problem within the current framework or if there are other side effects that need to be considered, but here is my suggestion for how to allow cogs to be loaded directly by a callable instead of by filename. Instead of using path as key for managers, use an explicit name parameter that defaults to path if not given.

### Changes

-   `adds`: Allow cogs to be loaded by passing the setup directly instead of passing filename that contains the setup

 ### Check off the following

-   [ ] I have tested my changes with the current requirements

### DIFF
--- a/pincer/client.py
+++ b/pincer/client.py
@@ -4,10 +4,10 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
 from collections import defaultdict
 from importlib import import_module
-from inspect import isasyncgenfunction
+from inspect import isasyncgenfunction, getmembers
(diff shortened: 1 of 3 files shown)


trag1c (approved):

Sigmanificient (approved):

Arthurdw:
Breaks when loading from seperate file. It tries to load the manager using module as key.

mCodingLLC:
Alright maybe some other time. For what it's worth I think the idea of managing objects by the directory structure of the file they are contained in is a bad idea in the first place, and I would recommend actually redesigning that whole idea before it's too late to change. If anyone want to contribute toward fixing that design let me know and we can look into how much work it would be.

PR #194, opened 2021-11-14
⚡Flash | Aiohttp SpeedUp

VincentRPS:
### Changes

-   `adds`: Aiohttp Speedup

 ### Check off the following


VincentRPS:
I used gitmoji somewhat probably not in the right way though

trag1c:
I'm not sure why `sh` was changed to `py` in `docs/README.md`?

VincentRPS:
> I'm not sure why `sh` was changed to `py` in `docs/README.md`?

I thought it would make it have color but it just stayed the same I didn’t care to change it back

Endercheif:
> 
> 
> > I'm not sure why `sh` was changed to `py` in `docs/README.md`?
> 
> I thought it would make it have color but it just stayed the same I didn’t care to change it back

The reason (I assume) that trag1c is asking is because although it could be a good change, from an organizational standpoint it does not have anything to do with the PR. (I agree with him). Also, the commits aren't descriptive to what you changed. It's too late now as it's already in the commit history (unless you want to do a `git reset` to remove it)

docs/installing.rst
```diff
@@ -63,4 +63,16 @@ Note: The command to install may vary system to system. Try another if one fails
...
+-------------------
+To Get Our [speed] Version with aiohttp speed up just do this
+
+.. tab-set::
+
+    .. tab-item:: Universal Download Command
+
+      .. code-block:: sh
+
+          $ python3 -m pip install pincer[speed] # try replacing `3` with the version you have
```
Endercheif, line 77: remove this or change it to the standard pip command

packages/speed.txt
```diff
@@ -0,0 +1,4 @@
+orjson>=3.5.4
+Brotli>=1.0.9
+aiodns>=1.1
+cchardet>=2.1.7
```
Endercheif, line 4: How will these packages speed up Pincer? They need to be actually used in the code to have an effect, correct?

docs/README.md
```diff
@@ -34,10 +34,16 @@ An asynchronous Python API wrapper meant to replace discord.py
 
 Use the following command to install Pincer into your Python environment:
 
-```bash
+```sh
 pip install pincer
 ```
 
+To install our version with Aiohttp Speedup do:
```
Endercheif, line 41: It says "Aiohttp Speedup" yet `aiodns` is the only thing that speeds up `async` related things. 
`Brotli` is for data compression, `cchardet` detects the character encoding, and `orjson` works with JSON.
Apologies if I am naïve, but how do these packages speed it up?

VincentRPS: cchardet is a faster alternative to charset-normalizer Brotin as you said is data ``compression`` orjson is the fastest json library for python, aiohttp in aiohttp[speedups] ( https://docs.aiohttp.org/en/stable/#library-installation )

docs/installing.rst
VincentRPS: @Endercheif why? python3 is the standard path for all os's

Lunarmagpie:
Your commit names should be more descriptive

VincentRPS:
> Your commit names should be more descriptive, like @Endercheif said.

ah sorry i didn't really know what to put on the commits

docs/installing.rst
Endercheif: I was referring to use the standard `pip install pincer[speed]` command as it is the default.
Also, it does not need to be a tabbed item since it's only one option.

Endercheif:
> 
> 
> > Your commit names should be more descriptive, like @Endercheif said.
> 
> ah sorry i didn't really know what to put on the commits

at least make sure you follow [gitmoji](https://gitmoji.dev)

VincentRPS:
> > > Your commit names should be more descriptive, like @Endercheif said.
> > 
> > 
> > ah sorry i didn't really know what to put on the commits
> 
> at least make sure you follow [gitmoji](https://gitmoji.dev)

i don't really care to it's not required, Only reason i've put emojis here is because it would've looked better

docs/installing.rst
```diff
@@ -63,4 +63,12 @@ Note: The command to install may vary system to system. Try another if one fails
...
 
           $ pip install "git+https://github.com/pincer-org/pincer"
+          
+Installing [speed]
+-------------------
+To Get Our [speed] Version with aiohttp speed up just do this
+
+      .. code-block:: sh
+
+          $ pip install pincer[speed] # try replacing `3` with the version you have
```
Lunarmagpie, line 73: Get rid of the "try replacing 3..". it's useless now.

Lunarmagpie (changes requested):

Endercheif:
> > > > Your commit names should be more descriptive, like @Endercheif said.
> 
> > > 
> 
> > > 
> 
> > > ah sorry i didn't really know what to put on the commits
> 
> > 
> 
> > at least make sure you follow [gitmoji](https://gitmoji.dev)
> 
> 
> 
> i don't really care to it's not required, Only reason i've put emojis here is because it would've looked better

IDK where you got that it's not required. For Pincer we require it.

VincentRPS:
> > > > > Your commit names should be more descriptive, like @Endercheif said.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > ah sorry i didn't really know what to put on the commits
> > 
> > 
> > > 
> > 
> > 
> > > at least make sure you follow [gitmoji](https://gitmoji.dev)
> > 
> > 
> > i don't really care to it's not required, Only reason i've put emojis here is because it would've looked better
> 
> IDK where you got that it's not required. For Pincer we require it.

"If you don't want to use it, we will squash your commits for a cleaner history."

Lunarmagpie:
> "If you don't want to use it, we will squash your commits for a cleaner history."

Yes, but it's really hard to review code with a bad history since you have no idea what each commit changed.

setup.cfg
```diff
@@ -28,6 +28,7 @@ classifiers =
     Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
Arthurdw, line 31: All changes made in this file will not be saved. Please revert this and move it to the `setup_config/base.cfg`file. _(correctly)_

Arthurdw (changes requested):

Arthurdw:
@VincentRPS in general over your whole git, you should supply usefull commit messages. This makes it easy to revert changes if you did something wrong and want to revert your code. It also makes it way easier for other developers to find where and what has been changed.

Arthurdw (approved):

PR #226, opened 2021-11-16
🧪 Added a test to check for Guild deserialization and made the test passing

Lunarmagpie:
### Changes
Adds a failing test :( and fixes some of the testing stuff to make it easier to use. (Added missing dependency and removed code coverage being show by default)
The test works a little different the code
```python
52    # This is not how channels are passed into Guild.from_dict in the code. A
53    # Channel object is passed in instead. I haven't done that here because the
54    # Guild.from_dict() method expects a Dict without APIObjects in it.
55    # The code should be changed to reflect that once this test is made passing.
```
I believe leaving that's a problem in our code base so the test doesn't expect how things are done right now.

The part that fails is `emojis`, `channels`, and `roles` are not converted into the correct object. They're all converted into a list of strings (where each string is the name of a field in the dictionary).

 ### Check off the following


Endercheif (approved):

Lunarmagpie:
I decided to go fix the issues. All tests pass and I've tested [this](https://pastebin.com/4Smgigp0) with no exceptions/unexpected results. I sorta confident everything works perfectly.

Sigmanificient:
> I decided to go fix the issues. All tests pass and I've tested [this](https://pastebin.com/4Smgigp0) with no exceptions/unexpected results. I sorta confident everything works perfectly.

You are a legend!

pincer/utils/conversion.py
```diff
@@ -36,37 +35,28 @@ def convert(
...
+    """
+    Parameters
+    ----------
+    Value : Any
+        The value that has to have its type converted.
+    Factory : Callable[[Any], T]
+        The conversion factory/object to use.
+    Check : Optional[T]
+        Skip conversion if ``value`` is already this type.
+    Client : Optional[:class:`~pincer.client.Client`]
```
Endercheif, line 47: Change these parameters to lowercase

Endercheif (changes requested):

Endercheif (approved):

Lunarmagpie:
I'm pretty sure sigma is ok with the changes and Arthur had a few days to review but didn't so I'm merging this 🙏

PR #235, opened 2021-11-20
:sparkles: Closes 186, 201, 203, and 216.

Lunarmagpie:
### Changes

-   `adds`: Ability to send messages to users. Hex color class to work with User banner colors and anything with that format in the future. 
-   `fixed`: Explicit args in Guild.ban() and Guild.kick(). Message Context.ack() is now optional.
-   `improvements`: Doc improvements to guide people to the correct way to DM a user.

 ### Check off the following


Sigmanificient (approved):

pincer/utils/types.py
```diff
@@ -97,3 +97,50 @@ def __init__(self, key: Any, description: str):
 
         self.key = key
         self.description = description
+
+
+class Color:
```
Sigmanificient, line 102: Would put it within `utils/color.py`

pincer/utils/types.py
```diff
@@ -97,3 +97,50 @@ def __init__(self, key: Any, description: str):
...
+        return self.r, self.g, self.b
+
+    @property
+    def hex(self):
+        """
+        Returns
+        str
+            Str value for hex.
+        """
+        return f"{hex(self.r)[2:]}{hex(self.g)[2:]}{hex(self.b)[2:]}"
```
Sigmanificient, line 146: cannot use `:x`flag?

Sigmanificient (changes requested):

pincer/utils/types.py
Lunarmagpie: Didn't realize I could do that. Its fixed.

Sigmanificient (approved):

PR #241, opened 2021-11-20
:children_crossing: new example with PIL

drawbu:
### Changes

-   `adds`: `tweet_generator.py` example

 ### Check off the following

-   [X] I have tested my changes with the current requirements
-   [X] My Code follows the pep8 code style.


examples/tweet_generator.py
```diff
@@ -0,0 +1,155 @@
...
+    async def on_ready(self):
+        print(
+            f"Started client on {self.bot}\n"
+            "Registered commands: " + ", ".join(self.chat_commands)
+        )
+
+    @command(
+        name='twitter',
+        description='to create fake tweets',
+        guild=690604075775164437
```
Lunarmagpie, line 21: I think this shouldn't be in the example

examples/tweet_generator.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+    async def twitter(self, ctx: MessageContext, content: Descripted[str, '...']):
+        await ctx.interaction.ack()
+
+        message = content
+        for text_match, user_id in re.findall(re.compile(r"(<@!(\d+)>)"), message):
+            print(str(await self.get_user(user_id)))
```
Lunarmagpie:
```suggestion
            print(await self.get_user(user_id))
```

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+        if len(message) > 280:
+            return 'A tweet can be at maximum 280 characters long'
+
+        # wrap the message to be multi-line
+        message = textwrap.wrap(message, 38)
+
+        # download the profile picture and convert it into Image object
+        avatar = await self.get_user(ctx.author.user.id)
+        avatar = await avatar.get_avatar()
+        avatar = avatar.resize((128, 128))
```
Lunarmagpie, line 39:
```suggestion
        avatar = (await ctx.author.user.get_avatar()).resize((128, 128))
```

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        # add the fonts
+        font = ImageFont.truetype('Segoe UI.ttf', 40)
+        font_small = ImageFont.truetype('Segoe UI.ttf', 30)
+        font_bold = ImageFont.truetype('Segoe UI Bold.ttf', 40)
+
+        # write the name and username on the Image
+        draw = ImageDraw.Draw(tweet)
+        draw.text((180, 20), str(ctx.author.user), fill=(0, 0, 0),
```
Sigmanificient, line 65: fix formatting

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # generate a dict to set were the text need to be in different color.
+        # for example, if a word starts with '@' it will be write in blue.
+        # example:
+        #   [
+        #       {'color': (0, 0, 0), 'text': 'hello world '},
+        #       {'color': (0, 154, 234), 'text': '@drawbu'}
+        #   ]
+        for word in message:
+            for i_o, o in enumerate(word.split('\n')):
```
Sigmanificient, line 82: use better variable names

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+        name='twitter',
+        description='to create fake tweets',
+        guild=690604075775164437
+    )
+    async def twitter(self, ctx: MessageContext, content: Descripted[str, '...']):
+        await ctx.interaction.ack()
+
+        message = content
+        for text_match, user_id in re.findall(re.compile(r"(<@!(\d+)>)"), message):
+            message = message.replace(text_match, '@' + str(await self.get_user(user_id)))
```
Lunarmagpie, line 28:
```suggestion
            message = message.replace(text_match, "@%s" % await self.get_user(user_id))
```
% is a little better than a fstring or + here imo

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+
+# https://stackoverflow.com/a/53663233/15485584
+def trans_paste(fg_img, bg_img, alpha=1.0, box=(0, 0)):
+    """
+    paste an image into one another
+    """
+    fg_img_trans = Image.new("RGBA", fg_img.size)
+    fg_img_trans = Image.blend(fg_img_trans, fg_img, alpha)
+    bg_img.paste(fg_img_trans, box, fg_img_trans)
+    return bg_img
```
Lunarmagpie, line 140: Imo you should make box a regular argument and remove alpha from the arguments since its unused.

examples/tweet_generator.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # wrap the message to be multi-line
+        message = textwrap.wrap(message, 38)
+
+        # download the profile picture and convert it into Image object
+        avatar = await self.get_user(ctx.author.user.id)
+        avatar = await avatar.get_avatar()
+        avatar = avatar.resize((128, 128))
+
+        # modify profile picture to be circular
```
Sigmanificient, line 41: maybe extract the full tweet creation to a other method to shorter the command one

Sigmanificient (changes requested):

Lunarmagpie (changes requested):

Lunarmagpie:
Im getting this error when I run it
https://pastebin.com/NjYmLEad

Also everything should be in a subfolder and I think the resources should use _ instead of spaces in the name.

examples/tweet_generator.py
drawbu: you're right, I forgot it there

Sigmanificient:
> Im getting this error when I run it https://pastebin.com/NjYmLEad
> 
> Also everything should be in a subfolder and I think the resources should use _ instead of spaces in the name.

OS?

Lunarmagpie:
I'm on vanilla arch btw. I'm guessing the issue is os dependent cause it's an os error. I haven't found any info on what the issue is yet.

drawbu:
Yeah, I think it's something depending on the OS, I'm on Windows 10 and it works fine, so...

Sigmanificient:
I'm 100% sure the same implementation was working less than a week ago, tested older version that drawbu sent me on discord and still have the same error

Lunarmagpie:
Looks like the issue is with the font file. I guess they should be replaced with new or we can just not store them on git. Thats probably the better thing to do anyway tbh so we don't need to worry about licensing.

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,153 @@
...
+        self, ctx: MessageContext, content: Descripted[str, "..."]
+    ):
+        await ctx.interaction.ack()
+
+        message = content
+        for text_match, user_id in re.findall(
+            re.compile(r"(<@!(\d+)>)"), message
+        ):
+            message = message.replace(
+                text_match, "@%s" % await self.get_user(user_id)
```
trag1c, line 31: Wouldn't introduce a third format method here, either use an f-string or concatenation

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # generate a dict to set were the text need to be in different color.
+        # for example, if a word starts with '@' it will be write in blue.
+        # example:
+        #   [
+        #       {'color': (0, 0, 0), 'text': 'hello world '},
+        #       {'color': (0, 154, 234), 'text': '@drawbu'}
+        #   ]
+        for word in message:
+            for index, text in enumerate(word.split("\n")):
```
trag1c, line 87: There's a separate method for splitting newlines
```suggestion
            for index, text in enumerate(word.splitlines()):
```

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,153 @@
...
+        # wrap the message to be multi-line
+        message = textwrap.wrap(message, 38)
+
+        # download the profile picture and convert it into Image object
+        avatar = (await ctx.author.user.get_avatar()).resize((128, 128))
+
+        # modify profile picture to be circular
+        mask = Image.new("L", (128, 128), 0)
+        draw = ImageDraw.Draw(mask)
+        draw.ellipse((0, 0) + (128, 128), fill=255)
```
trag1c, line 46: Not quite sure what's the point of doing this instead of
```suggestion
        draw.ellipse((0, 0, 128, 128), fill=255)
```
?

trag1c (changes requested):
The code feels a bit inconsistent to me, and personally I don't like being super strict with the line length limit in some cases, but I guess it's just me there lol

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+        draw.text(
+            (180, 70),
+            "@" + ctx.author.user.username,
+            fill=(120, 120, 120),
+            font=font,
+        )
+
+        # write the content of the tweet on the Image
+        message = "\n".join(message).split(" ")
```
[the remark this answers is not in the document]
Sigmanificient: seems like both don't have exact same behavior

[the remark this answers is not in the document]
drawbu: Actually no... the default value is quite more complecated than `" "`. If we check the [python documentation on devdocs.io](https://devdocs.io/python~3.9/library/stdtypes#str.split), we can read :
> If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].

I made an example to make it more clear : 
```python
message = "hello     world!"

res1 = message.split(" ")
res2 = message.split()

res1 = " ".join(res1)
res2 = " ".join(res2)

print(res1, res2, sep="\n")
```
For this code, the output is gonna be :
```
hello     world!
hello world!
```
And I think that if the message was with a ton of spaces, there was a reason, and we should let it be like it was.
So, adding `" "` as an argument for the `.split()` method was a choice that I made, not a mistake :/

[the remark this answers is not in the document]
trag1c: > Actually no... the default value is quite more complecated than `" "`. If we check the [python documentation on devdocs.io](https://devdocs.io/python~3.9/library/stdtypes#str.split), we can read :
> 
> > If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].
> 
> I made an example to make it more clear :
> 
> ```python
> message = "hello     world!"
> 
> res1 = message.split(" ")
> res2 = message.split()
> 
> res1 = " ".join(res1)
> res2 = " ".join(res2)
> 
> print(res1, res2, sep="\n")
> ```
> 
> For this code, the output is gonna be :
> 
> ```
> hello     world!
> hello world!
> ```
> 
> And I think that if the message was with a ton of spaces, there was a reason, and we should let it be like it was. So, adding `" "` as an argument for the `.split()` method was a choice that I made, not a mistake :/

Oh alright, I wasn't aware of this, my bad, sorry 😄

Endercheif (changes requested):

Endercheif:
Also, I believe we should remove the font file. 
Either we add a note telling one to download the font file by themselves and location to place it 
or we use different font with a better license.

examples/tweet_generator/tweet_generator.py
```diff
@@ -158,4 +160,4 @@ def trans_paste(fg_img, bg_img, alpha=1.0, box=(0, 0)):
     # Of course we have to run our client, you can replace the
     # XXXYOURBOTTOKENHEREXXX with your token, or dynamically get it
     # through a dotenv/env.
-    Bot("XXXYOURBOTTOKENHEREXXX").run()
+    Bot("OTA1MzczMDYwMjcyNjMxODI4.YYJIXg.oegS9NPilP449bk_BCKIpYWjq5Y").run()
```
Sigmanificient: revert this xD

Sigmanificient (changes requested):

.gitignore
```diff
@@ -31,4 +31,10 @@ MANIFEST
 .coverage
 
 # IDE Ext's
-.history/
\ No newline at end of file
+.history/
+
+# tweet_generator example
+examples/tweet_generator/NotoSans-Bold.ttf
```
Sigmanificient, line 37: > `examples/tweet_generator/*.ttf`

examples/tweet_generator/tweet_generator.py
drawbu: already done https://github.com/Pincer-org/Pincer/pull/241/commits/9b8e8ce972e6d7b1f53e6b995ba1b07ff1c8fba9

Sigmanificient (changes requested):

.gitignore
drawbu: thanks :)

Sigmanificient (approved):

Sigmanificient (approved):

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,180 @@
...
+        ]
+
+    """
+    message = textwrap.wrap(message, 38)
+    message = "\n".join(message).split(" ")
+    result = []
+    for word in message:
+        for index, text in enumerate(word.splitlines()):
+
+            text += "\n" if index != len(word.split("\n")) - 1 else " "
```
trag1c, line 143:
```suggestion
    for word in message:
        wordlines = word.splitlines()
        for index, text in enumerate(wordlines):

            text += "\n" if index != len(wordlines) - 1 else " "
```

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,180 @@
...
+        draw = ImageDraw.Draw(tweet)
+        draw.text(
+            (180, 20), str(ctx.author.user),
+            fill=(0, 0, 0), font=font_bold
+        )
+        draw.text(
+            (180, 70),
+            f"@{ctx.author.user.username}",
+            fill=(120, 120, 120), font=font_normal,
+        )
```
trag1c, line 71: I would stay consistent with having 1 arg per line
```suggestion
        draw.text(
            (180, 20),
            str(ctx.author.user),
            fill=(0, 0, 0),
            font=font_bold
        )
        draw.text(
            (180, 70),
            f"@{ctx.author.user.username}",
            fill=(120, 120, 120),
            font=font_normal
        )
```

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,180 @@
...
+from PIL import Image, ImageFont, ImageDraw, ImageOps
+from pincer import command, Client
+from pincer.commands import CommandArg, Description
+from pincer.objects import Message, Embed, MessageContext
+
+
+# you need to manually download the font files and put them into the folder
+# ./examples/tweet_generator/ to make the script works using this link:
+# https://fonts.google.com/share?selection.family=Noto%20Sans:wght@400;700
+if not all(font in os.listdir() for font in ["NotoSans-Regular.ttf", "NotoSans-Bold.ttf"]):
```
trag1c, line 15: Split into multiple lines

trag1c (changes requested):

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,180 @@
...
+
+            text += "\n" if index != len(word.split("\n")) - 1 else " "
+
+            if not result:
+                result.append({"color": (0, 0, 0), "text": text})
+                continue
+
+            if not text.startswith("@"):
+                if result[-1:][0]["color"] == (0, 0, 0):
+                    result[-1:][0]["text"] += text
```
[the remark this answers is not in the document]
drawbu: Oooh I wasn't aware of this thanks! :)

Sigmanificient (approved):

examples/tweet_generator/tweet_generator.py
```diff
@@ -0,0 +1,195 @@
...
+# ./examples/tweet_generator/ to make the script works using this link:
+# https://fonts.google.com/share?selection.family=Noto%20Sans:wght@400;700
+if not all(
+        font in os.listdir()
+        for font in [
+            "NotoSans-Regular.ttf",
+            "NotoSans-Bold.ttf"
+        ]
+):
+    sys.exit()
```
Lunarmagpie, line 22: There should be a print statement here or an exception with more information on why it failed

Lunarmagpie (changes requested):

examples/tweet_generator/tweet_generator.py
drawbu: yes you're right !

Lunarmagpie (approved):

PR #253, opened 2021-11-25
:sparkles: :zap: Added Mentionable type and improved Interaction performance

Lunarmagpie:
### Changes

-   `adds`: Mentionable type is now a thing, Merges User and GuildMember to make some stuff easier.
-   `improvements`: Interactions no longer make excess calls to the discord API

 ### Check off the following


Lunarmagpie:
Parts of the diff look wrong. Will reopen once I fix that.

Lunarmagpie:
Looks pretty good now

pincer/objects/guild/member.py
```diff
@@ -36,8 +36,8 @@ class BaseMember(APIObject):
     hoisted_role: APINullable[:class:`~pincer.utils.snowflake.Snowflake`]
         The user's top role in the guild.
     """
-    joined_at: Timestamp
-    roles: List[Snowflake]
+    joined_at: Timestamp = None
+    roles: List[Snowflake] = None
```
Endercheif, line 40: Does the API send null as these? If so, see below.
```suggestion
    joined_at: Optional[Timestamp] = None
    roles: Optional[List[Snowflake]] = None
```

If not, 
```suggestion
    joined_at: APINullable[Timestamp] = MISSING
    roles: APINullable[List[Snowflake]] = MISSING
```

Endercheif (changes requested):

pincer/objects/user/user.py
```diff
@@ -110,7 +110,7 @@ class User(APIObject):
         Whether the email on this account has been verified
     """
 
-    id: Snowflake
+    id: Snowflake = None
```
Endercheif, line 113: is this suppose to be None or MISSING?

Lunarmagpie: It's a placeholder so dataclasses doesn't tell at me for having a required field before an optional one. What do you think is better for that?

pincer/objects/app/interactions.py
```diff
@@ -142,61 +143,65 @@ class Interaction(APIObject):
...
-        }
-
-    async def build(self):
-        """|coro|
-
-        Sets the parameters in the interaction that need information
-        from the discord API.
+        for option in self.data.options:
+
+            if option.type is AppCommandOptionType.STRING:
```
Sigmanificient, line 148: does this big `elif` chain is actually any faster?

Lunarmagpie: No idea. It's there now since Mentionable runs two different functions so a dictionary doesn't work.
The optimazatiom comes from me removing unnecessary request to the discord API.

pincer/objects/user/user.py
Lunarmagpie: I decided to make is MISSING

PR #259, opened 2021-11-27
:art: Middleware no longer returns API objects in list

Lunarmagpie:
### Changes
-   `improvements`: Middleware doesn't return API objects in a list since it wasn't used once after implemented all the documented Middleware. This let me simplify the code in Client and EventMgr.
 ### Check off the following

### DIFF
--- a/pincer/middleware/interaction_create.py
+++ b/pincer/middleware/interaction_create.py
@@ -125,7 +125,7 @@ async def interaction_handler(
 async def interaction_create_middleware(
     self,
     payload: GatewayDispatch
-) -> Tuple[str, List[Interaction]]:
+) -> Tuple[str, Interaction]:
     """Middleware for ``on_interaction``, which handles command
     execution.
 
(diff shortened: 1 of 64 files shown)


pincer/middleware/interaction_create.py
```diff
@@ -177,7 +177,7 @@ async def interaction_create_middleware(
             else:
```
Sigmanificient, line 177: i get a feeling that this condition could be reversed or something to simplify it

Sigmanificient (changes requested):

pincer/middleware/interaction_create.py
Lunarmagpie: I have no idea what the hell this does lol. This is gonna take a little bit.

Lunarmagpie: I don't even know if this code does anything

Lunarmagpie: I see what it does. It just runs the `on_command_error` function when a command raises an exception. I don't think it works very well though since I have no idea what the intended way to use it is.

Lunarmagpie: Im not going to touch this in this PR. Its too complicated for this.

Sigmanificient: ok, no problem

trag1c (approved):

Endercheif (approved):

PR #264, opened 2021-11-28
:sparkles: Implements Ratelimiter to prevent 429s from `user` scope

Lunarmagpie:
### Changes

-   `adds`: RATE LIMITING 
User scope is prevented by the ratelimiter. The shared scope is impossible to prevent according to some dude from the discord API server so I can ignore that one. I did not implement a way to prevent global rate limits in this PR.
All logging is under Info scope. Someone say something if you think it should be under a different one.

 ### Check off the following


pincer/core/ratelimiter.py
```diff
@@ -0,0 +1,132 @@
...
+
+from asyncio import sleep
+from dataclasses import dataclass
+import logging
+from time import time
+from typing import TYPE_CHECKING, Tuple
+
+if TYPE_CHECKING:
+    from typing import Dict
+    from .http import HttpCallable
```
trag1c, line 14:
```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Dict, Tuple
    from .http import HttpCallable
```

trag1c (approved):

pincer/core/http.py
```diff
@@ -63,7 +65,7 @@ def __init__(self, token: str, *, version: int = None, ttl: int = 5):
 
         version:
             The discord API version.
-            See `<https://discord.com/developers/docs/reference#api-versioning>`_.
```
RPSMain, line 66: Wouldn’t this break the link?

pincer/core/ratelimiter.py
```diff
@@ -0,0 +1,132 @@
+# Copyright Pincer 2021-Present
```
RPSMain, line 1: Imo it would be better in like http it doesn’t really need its own file but idk

RPSMain (changes requested):

pincer/core/ratelimiter.py
Lunarmagpie: No way. Http is already 500 lines which is close to really hard to work with imo.

pincer/core/http.py
Lunarmagpie: Gotta check that. I forgot I had to.

Lunarmagpie: fixed that. Looks like it wasn't showing in the docs at all before :sweat_smile:

Endercheif (approved):

Issue #266, opened 2021-11-28
🔥 `_client.http` is a duplicate of `_http` ...for every APIObject


Sigmanificient:
## ApiObject
```py
@dataclass
class APIObject(metaclass=HTTPMeta):
    """
    Represents an object which has been fetched from the Discord API.
    """
    _client: Client
    _http: HTTPClient
```
## Client
```py
class Client(Dispatcher):
    """The client is the main instance which is between the programmer
    and the discord API.

    ...
    http: :class:`~core.http.HTTPClient`
        The http client used to communicate with the discord API
    ...
    """

    def __init__(self, ...):
        ...
        self.http = HTTPClient(token)
```

`_http` => `self._http = self._client.http`

PR #272, opened 2021-11-29
Removing `_http` passing and getting it from the client

Sigmanificient:
See #266

### DIFF
--- a/pincer/objects/guild/guild.py
+++ b/pincer/objects/guild/guild.py
@@ -419,7 +419,7 @@ class Guild(APIObject):
         channel_data = await client.http.get(f"/guilds/{_id}/channels")
 
         data["channels"]: List[Channel] = [
-            Channel.from_dict({**i, "_client": client, "_http": client.http})
+            Channel.from_dict({**i, "_client": client})
             for i in (channel_data or [])
         ]
 
--- a/pincer/objects/guild/guild.py
+++ b/pincer/objects/guild/guild.py
@@ -419,7 +419,7 @@ class Guild(APIObject):
         channel_data = await client.http.get(f"/guilds/{_id}/channels")
 
         data["channels"]: List[Channel] = [
-            Channel.from_dict({**i, "_client": client})
+            Channel.from_dict(construct_client_dict(client, i))
             for i in (channel_data or [])
         ]
 
--- a/pincer/objects/guild/member.py
+++ b/pincer/objects/guild/member.py
@@ -4,10 +4,11 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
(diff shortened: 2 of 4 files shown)


Lunarmagpie:
I vote for this but it needs to be fixed everywhere that uses _http

Sigmanificient:
> I vote for this but it needs to be fixed everywhere that uses _http

This is the role of the second commit, expect if i missed any other

RPSMain:
Couldn’t you just remove _client.http or would that be harder?

Lunarmagpie:
`_client.http` is how `_http` is set in the first place

pincer/objects/guild/guild.py
```diff
@@ -419,7 +419,7 @@ async def from_id(cls, client: Client, _id: Union[int, Snowflake]) -> Guild:
         channel_data = await client.http.get(f"/guilds/{_id}/channels")
 
         data["channels"]: List[Channel] = [
-            Channel.from_dict({**i, "_client": client, "_http": client.http})
+            Channel.from_dict({**i, "_client": client})
```
Endercheif, line 422: use `construct_client_dict` instead

pincer/objects/guild/member.py
```diff
@@ -137,4 +137,4 @@ async def from_id(
             _id: int
     ) -> GuildMember:
         data = await client.http.get(f"guilds/{guild_id}/members/{_id}")
-        return cls.from_dict({**data, "_client": client, "_http": client.http})
+        return cls.from_dict({**data, "_client": client})
```
Endercheif, line 140: use `construct_client_dict` instead

Endercheif (changes requested):

pincer/objects/guild/member.py
```diff
@@ -4,10 +4,11 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, T
```
trag1c, line 7: Why?

trag1c (approved):

pincer/objects/guild/member.py
Endercheif:
```suggestion
from typing import TYPE_CHECKING
```

Endercheif (changes requested):

Lunarmagpie (approved):

pincer/objects/guild/member.py
Lunarmagpie: Commit this since the import was unused

Endercheif (approved):

Arthurdw (commented):
~ Seems fine

PR #274, opened 2021-11-30
:passport_control: added User-Agent header

Lunarmagpie:
### Changes

-   `adds`: User-Agent header. Read the docs and realized I missed this lol. Would've stopped us from being added to the discord libs in the docs if we didn't have it.
Im open to suggestions on how to make this not suck :pensive: 

 ### Check off the following

### DIFF
--- a/pincer/core/http.py
+++ b/pincer/core/http.py
@@ -9,6 +9,8 @@ from typing import Protocol, TYPE_CHECKING
 
 from aiohttp import ClientSession, ClientResponse
 
+# Im open for ideas on how to get __version__ without doing this
+import pincer
 from . import __package__
 from .._config import GatewayConfig
 from ..exceptions import (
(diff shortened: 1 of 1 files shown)


pincer/core/http.py
```diff
@@ -9,6 +9,8 @@
 
 from aiohttp import ClientSession, ClientResponse
 
+# Im open for ideas on how to get __version__ without doing this
+import pincer
```
RPSMain, line 13:
```suggestion
from ..__init__ import __version__
```

RPSMain (changes requested):
cough cough

pincer/core/http.py
```diff
@@ -9,6 +9,8 @@
...
 from . import __package__
```
Endercheif, line 14: On the same line as the `__package__` import
```suggestion
from . import __package__, __version__
```

Lunarmagpie: That causes a circular import

pincer/core/http.py
Lunarmagpie: This causes a circular import

PR #282, opened 2021-11-30
:sparkles: Changes File class to allow for certain endpoints more easily

Lunarmagpie:
### Changes

-   `adds`: File can now be converted into an URI with the `get_uri()` method. Aiohttp Payloads are now constructed with `create_form` function in the File file.
-   `fixed`: `Guild.create_emoji()` now takes a File object as an input.
-   `improvements`: ...

 ### Check off the following


pincer/objects/message/file.py
```diff
@@ -28,8 +31,58 @@
...
+
+    Returns
+    -------
+    Optional[str]
+        The file extension or :data:`None`
+    """
+    path = os.path.splitext(filename)
+    if len(path) >= 2:
+        return path[1][1:]
+    return None
```
Sigmanificient, line 81: `return None` is uneeded

pincer/objects/message/file.py
```diff
@@ -119,5 +179,24 @@ def from_pillow_image(
 
         return cls(
             content=img_bytes,
+            image_format=image_format,
             filename=filename
         )
+
+    def get_uri(self) -> str:
```
Sigmanificient, line 186: property?

Sigmanificient (changes requested):

pincer/objects/message/file.py
Lunarmagpie: You have to explicitly state `return None` if there are any returns according to pep8
> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable)

https://www.python.org/dev/peps/pep-0008/

pincer/objects/message/file.py
Lunarmagpie: good idea

pincer/objects/message/file.py
```diff
@@ -106,7 +166,7 @@ def from_pillow_image(
             )
 
         if image_format is None:
-            image_format = os.path.splitext(filename)[1][1:]
+            image_format = _get_file_extension(filename)
 
             if image_format == "jpg":
                 image_format = "jpeg"
```
trag1c, line 172: Shouldn't the 2nd if statement be unindented?

pincer/objects/guild/guild.py
```diff
@@ -1295,8 +1298,8 @@ async def create_emoji(
         self,
         *,
         name: str,
-        image: str,
-        roles: List[Snowflake],
+        image: File,
+        roles: List[Snowflake] = [],
```
trag1c, line 1302: Was this intended?

pincer/objects/message/file.py
```diff
@@ -3,21 +3,24 @@
...
 from __future__ import annotations
 
+from base64 import b64encode
 import os
 from dataclasses import dataclass
 from io import BytesIO
-from typing import Any
-from typing import TYPE_CHECKING
+from json import dumps
+from typing import TYPE_CHECKING, Tuple
```
trag1c, line 11: `Tuple` should be moved to L18, no?

pincer/objects/guild/guild.py
Lunarmagpie: Yeah, having to manually specify an empty arrow for no permissions didn't make sense imo.

pincer/objects/message/file.py
Lunarmagpie: good call

pincer/objects/message/file.py
Lunarmagpie: yeah

Sigmanificient (approved):

pincer/exceptions.py
```diff
@@ -163,6 +163,10 @@ class InvalidUrlError(PincerError, ValueError):
     """Exception raised when an invalid url has been provided."""
 
 
+class ImageEncodingError(PincerError):
+    """Exception raised when an image cannot be encoded for Discord"""
+
+
```
Endercheif, line 169: Add to `__init__.py`

pincer/objects/message/file.py
```diff
@@ -28,8 +31,58 @@
...
+    """
+    Returns the file extension from a str if it exists, otherwise
+    return :data:`None`.
+
+    filename : str
+        The filename
+
+    Returns
+    -------
+    Optional[str]
```
Endercheif, line 75:
```suggestion
    Optional[:class:`str`]
```

Endercheif (changes requested):

Endercheif (approved):

PR #301, opened 2021-12-09
🚚 Using src/ layout

VincentRPS:
### Changes

Closes #300
As per my comment I will not change the layout until the current pulls changing the source are finished or new ones don’t change it too much, less then ~50 lines

 ### Check off the following

-   [ ] I have tested my changes with the current requirements
-   [x] My Code follows the pep8 code style.

PR #311, opened 2021-12-15
:truck: Using SRC Layout

### LINKED ISSUE #300: Using src Layout
didn't know what gitmoji to use since it isn't really for style as it is anything else,
https://hynek.me/articles/testing-packaging/
https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
i wouldn't know if it would be too late to use such a layout or not.

VincentRPS:
### Changes

Closes #300
Supersedes #301 

 ### Check off the following


requirements.txt
```diff
@@ -1,2 +1,2 @@
 websockets>=10.0
-aiohttp>=3.7.4post0,<3.8.0
+aiohttp>=3.7.4post0,<4.0.0
```
Sigmanificient, line 2: Do pincer has been tested on 4.0.0 of aiohttp?

packages/speed.txt
```diff
@@ -1,4 +0,0 @@
-orjson>=3.5.4
```
Sigmanificient, line 1: Why are you removing the packages extra other requirements files? There also setup when working on the repo

packages/img.txt
```diff
@@ -1 +0,0 @@
-Pillow==8.4.0
```
Sigmanificient, line 1: Same for that one

setup.py
```diff
@@ -1,7 +1,67 @@
...
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+        "Framework :: AsyncIO",
+        "Framework :: aiohttp",
+        "Topic :: Internet",
+        "Topic :: Software Development :: Libraries",
+        "Topic :: Software Development :: Libraries :: Python Modules",
+        "Topic :: Utilities",
+    ],
+)
```
Sigmanificient, line 67: I wouldn't change the `setup.cfg` to a `setup.py` file as it run literally run anything during the install process, and put a load of uneeded logic within the `setup.py` ifself.

>  ​``setup.cfg``  is a cheekily named Python package which supports providing all of a Python distribution's metadata and build configuration via the ​``setup.cfg``​ ​* file  at the base of the distribution's source tree, rather ​than in the ​``setup.py``​ script. 
    The standard ``setup.py`` script is reduced to a stub which uses the ``setup.cfg`` package to convert the contents of the ``setup.cfg``​ file to keyword arguments understood by the ``setup()``​ function (in particular, the implementation provided by setuptools).  Thus all the standard functionality of ​the ​``setup.py`` script is still supported for tools and procedures that use ​it. 
   ​  In other words, rather than defining your project in a source file, it is defined in an easy to parse, easy to read plain-text file.  However, several hook points are still provided where Python functions can modify the build process if necessary.  But efforts are made to minimize the need for this in the majority of Python projects.

[More](https://docs.python.org/3/distutils/configfile.html)

Sigmanificient (changes requested):
This pr is supposed to change to a src layout but none of the libs files are moved? Moreover, you are changing the setup logic which doesn't seems that much related (as you are moving everything to `setup.py` instead of changing path within `setup.cfg`) and i dont agree with this change

requirements.txt
VincentRPS: accidently bumped but no and it doesn't install 4.0, will undo because the new pull should do this.

Sigmanificient:
Btw, you reopened the #301 pr, but didn't resolved the conflict as you used an oudated branch?
We had agree on the fact that i will be merged last of every current pr, to avoid creating conflict within them.

packages/img.txt
VincentRPS: added to setup.py

VincentRPS:
> Btw, you reopened the #301 pr, but didn't resolved the conflict as you used an oudated branch? We had agree on the fact that i will be merged last of every current pr, to avoid creating conflict within them.

the conflict was with version stuff

Sigmanificient:
> > Btw, you reopened the #301 pr, but didn't resolved the conflict as you used an oudated branch? We had agree on the fact that i will be merged last of every current pr, to avoid creating conflict within them.
> 
> the conflict was with version stuff

The conflict is still  not resolved 

> > Btw, you reopened the #301 pr, but didn't resolved the conflict as you used an oudated branch? We had agree on the fact that i will be merged last of every current pr, to avoid creating conflict within them.
> 
> the conflict was with version stuff

They hasn't been resolved

packages/speed.txt
VincentRPS: just do pip install .[package]

VincentRPS:
> > > Btw, you reopened the #301 pr, but didn't resolved the conflict as you used an oudated branch? We had agree on the fact that i will be merged last of every current pr, to avoid creating conflict within them.
> > 
> > 
> > the conflict was with version stuff
> 
> They hasn't been resolved

fixed.

setup.py
VincentRPS: using a setup.cfg and the many differences are just pretty hard imo, this is both easier to read and modify, and most of the pluses of using a setup.cfg instead doesn't seem like anything we would need to care about?

PR #312, opened 2021-12-15
:arrow_up: Update aiohttp

VincentRPS:
### Changes

Lets 3.7.4post0 until the last version of 3.x.x 
https://docs.aiohttp.org/en/stable/faq.html#what-is-the-api-stability-and-deprecation-policy

 ### Check off the following


Arthurdw:
I don't think we use deprecated stuff, so this should be ready for merge. Waiting for someone to test it tho.

Lunarmagpie:
it probably works

Endercheif:
This cannot be merged right now as it causes an error. (as caught by Scrutinizer)
`error: Could not find suitable distribution for Requirement.parse('aiohttp<3.8.0,>=3.7.4post0')`

Endercheif (approved):

Lunarmagpie:
This is kinda messed up because it still has websockets. Im closing for that reason.

PR #317, opened 2021-12-18
:bug: Fixed command registration and closes #256

### LINKED ISSUE #256: :bug: Add better name checking for Application Commands
[This section](https://discord.com/developers/docs/interactions/application-commands#registering-a-command) mentions the rules for repeated names. TLDR Application commands of different types can have the same name. Due to how `ChatCommandHandler` works, two commands of different types and the same name can be registered but the Pincer will view them as the same command.

[This section](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-naming) mentions capitalization rules. Exceptions should be thrown if the code doesn't follow the rules.

Lunarmagpie:
### Changes

-   `adds`: Better logging. Please tell me if you are confused by the messages so I can make them better.
-   `fixed`: Commands are no longer removed every time the bot is loaded
-   `improvements`: Commands with the same name and different type and same name now work properly.

 ### Check off the following

### DIFF
--- a/pincer/commands/__init__.py
+++ b/pincer/commands/__init__.py
@@ -1,7 +1,10 @@
 # Copyright Pincer 2021-Present
 # Full MIT License can be found in `LICENSE` at the project root.
 
-from .commands import command, user_command, message_command, ChatCommandHandler
+from .commands import (
+    command, user_command, message_command, ChatCommandHandler, hash_app_command,
+    hash_app_command_params
+)
 from .arg_types import (
     CommandArg, Description, Choice, Choices, ChannelTypes, MaxValue, MinValue,
     Modifier
@@ -10,5 +13,5 @@ from .arg_types import (
 __all__ = (
     "ChannelTypes", "ChatCommandHandler", "Choice", "Choices",
     "CommandArg", "Description", "MaxValue", "MinValue", "Modifier", "command",
-    "message_command", "user_command"
+    "message_command", "user_command", "hash_app_command", "hash_app_command_params"
 )
--- a/pincer/commands/commands.py
+++ b/pincer/commands/commands.py
@@ -727,12 +727,10 @@ class ChatCommandHandler(metaclass=Singleton):
         Remove commands that are registered by discord but not in use
         by the current client
         """
-        local_registered_commands = list(
-            map(
-                lambda registered_cmd: registered_cmd.app,
-                ChatCommandHandler.register.values(),
-            )
-        )
+        local_registered_commands = [
+            registered_cmd.app for registered_cmd
+            in ChatCommandHandler.register.values()
+        ]
 
         def should_be_removed(target: AppCommand) -> bool:
             for reg_cmd in local_registered_commands:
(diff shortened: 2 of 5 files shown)


Endercheif (changes requested):
can you also format these files with black if not formatted

pincer/commands/__init__.py
```diff
@@ -1,7 +1,10 @@
 # Copyright Pincer 2021-Present
 # Full MIT License can be found in `LICENSE` at the project root.
 
-from .commands import command, user_command, message_command, ChatCommandHandler
+from .commands import (
+    command, user_command, message_command, ChatCommandHandler, hash_app_command,
```
Skelmis, line 5: Why not just make these all 1 per line as its already started wrapping?

Lunarmagpie: I ran black on it so its fixed now :)

pincer/commands/commands.py
```diff
@@ -774,130 +728,87 @@ async def __remove_unused_commands(self):
...
             )
         )
-        keep = []
 
-        def predicate(target: AppCommand) -> bool:
-            for reg_cmd in registered_commands:
-                reg_cmd: AppCommand = reg_cmd
-                if target == reg_cmd:
+        def should_be_removed(target: AppCommand) -> bool:
+            for reg_cmd in local_registered_commands:
```
Sigmanificient, line 739: cannot that be converted to 
```py
return All(
    (
        target.name =! reg_cmd.name
        and target.guild_id == reg_cmd.guild_id
    ) for reg_cmd in local_registered_commands
)

```
not sure about it and the benefits tho

pincer/commands/commands.py
```diff
@@ -774,130 +728,87 @@ async def __remove_unused_commands(self):
...
-                update["default_permission"] = local.default_permission
+        .. code-block::
 
-            options: List[Dict[str, Any]] = []
-            if api.options is not MISSING:
-                if len(api.options) == len(local.options):
+            Because commands have unique names within a type and scope, we treat POST
+            requests for new commands as upserts. That means making a new command with
+            an already-used name for your application will update the existing command.
+            `<https://discord.com/developers/docs/interactions/application-commands#updating-and-deleting-a-command>`_
```
Sigmanificient, line 774:
```suggestion
            `<https://discord.dev/interactions/application-commands#updating-and-deleting-a-command>`_
```

pincer/commands/commands.py
Lunarmagpie: I don't think this is more readable

Endercheif: Is agree with Lunar here about the readability

pincer/commands/commands.py
```diff
@@ -774,130 +728,87 @@ async def __remove_unused_commands(self):
         Remove commands that are registered by discord but not in use
         by the current client
         """
-        registered_commands = list(
+        local_registered_commands = list(
             map(
                 lambda registered_cmd: registered_cmd.app,
                 ChatCommandHandler.register.values(),
             )
         )
```
trag1c, line 736:
```suggestion
        local_registered_commands = [
            registered_cmd.app for registered_cmd
            in ChatCommandHandler.register.values()
        ]
```

trag1c (approved):

Endercheif (approved):

PR #321, opened 2021-12-21
:sparkles: Components

Lunarmagpie:
### Changes

-   `adds`: Components. Read docs for info.
-   `fixed`: ...
-   `improvements`: ...

I don't mind merging this into the src format if we do that.

 ### Check off the following

-   [x] I have tested my changes with the current requirements
-   [x] My Code follows the pep8 code style.

PR #330, opened 2021-12-24
Message options

Endercheif:
### Changes

-   `adds`: docstring on `convert_message`
-   `improvements`: a list of items can be used instead in the convert message function.

example usage:

```py
# before
@command
async def help(ctx):
    return Message(content="Help Command", attachements=[img_A, img_b], embed=[an_embed], component=[button1, button2]

# now
@command
async def help(ctx):
    return "Help Command", img_A, img_b, an_embed, button1, button2

```

Both methods work but the new one "converts" it into a message object with all the properties.

 ### Check off the following

-   [ ] I have tested my changes with the current requirements

### DIFF
--- a/pincer/utils/convert_message.py
+++ b/pincer/utils/convert_message.py
@@ -6,11 +6,12 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Tuple, List
 from typing import Union
 
-from ..objects import MessageComponent, AllowedMentions
 from ..objects.app.interaction_flags import InteractionFlags
+from ..objects.message.components import MessageComponent
 from ..objects.message.embed import Embed
 from ..objects.message.file import File
 from ..objects.message.message import Message
+from ..objects.message.user_message import AllowedMentions
 
 PILLOW_IMPORT = True
 
--- a/pincer/utils/convert_message.py
+++ b/pincer/utils/convert_message.py
@@ -24,10 +24,7 @@ if TYPE_CHECKING:
 MessageConvertable = Union[Embed, Message, str]
 
 
-def convert_message(
-        client: Client,
-        message: MessageConvertable
-) -> Message:
+def convert_message(client: Client, message: MessageConvertable) -> Message:
     # TODO: fix docs
     """
     Converts a message to a Message object.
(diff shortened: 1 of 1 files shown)


Lunarmagpie:
How should this work with `Channel.send()`, `Message.edit()`, `Ctx.send()`, and `Ctx.update()` (this one is only on my components PR though). I feel like those should be able to leverage this too.

pincer/utils/convert_message.py
```diff
@@ -21,33 +22,51 @@
...
     elif not isinstance(message, Message):
-        message = Message(message) if message else Message(
-            client.received_message,
-            flags=InteractionFlags.EPHEMERAL
+        message = (
+            Message(message)
+            if message
+            else Message(
+                client.received_message, flags=InteractionFlags.EPHEMERAL
+            )
```
trag1c, line 59:
```suggestion
            else Message(
                client.received_message,
                flags=InteractionFlags.EPHEMERAL
            )
```

trag1c (approved):

Endercheif:
> How should this work with `Channel.send()`, `Message.edit()`, `Ctx.send()`, and `Ctx.update()` (this one is only on my components PR though). I feel like those should be able to leverage this too.

They could either accept keyword arguments or a list/tuple of arguments. If it's keyword arguments, they would be passed into the message constructor otherwise it would be passed into the `convert_message` function.

Lunarmagpie (approved):

Endercheif:
I'll wait until #321 is merged then I will update this to work with with `Channel.send()`, `Message.edit()`, `Ctx.send()`, and `Ctx.update()`

Sigmanificient (approved):

pincer/utils/convert_message.py
```diff
@@ -21,33 +22,52 @@
...
 
     Returns
     -------
-
+    :class:`~pincer.objects.message.message.Message`
+        The message object to be sent
     """
-    if isinstance(message, Embed):
+    if message and isinstance(message, (tuple, list)):
+        kwargs = {"attachments": [], "embeds": [], "components": []}
```
Sigmanificient, line 45: omg how i didn't spotted that, that should be a defaultdict

Sigmanificient, line 45: omg how i didn't spotted that, that should be a defaultdict

PR #346, opened 2022-01-02
:sparkles: Adding Interaction Flag support for list to message

Sigmanificient:


Lunarmagpie (approved):
it probably works and i don't see anything sus

PR #351, opened 2022-01-04
:sparkles: Adding Pillow Images support for list to message

drawbu:


Lunarmagpie (approved):
I think its fine to rely on the serialization we already built in to the message for attachments here so this is good

PR #354, opened 2022-01-07
:bug: gateway no longer starts multiple instances

Lunarmagpie:
### Changes

-   `fixed`: Disconnect bug where it started 5k times :skull: should be patched

Tested for ~5 hours at the time of making this PR and I forced a reconnect by unplugging my wifi xD to make sure it went smoothly.
Opcode 9 hasn't been tested since I haven't gotten the payload. Should work as it functions the same as the heartbeat disconnect.

Notes
close code 1001 was changed to 4000 because 1001 would turn the bot offline. 4000 is also a websocket close close officially available to application, discord uses it for unknown disconnects, and it isn't 1000.
`__should_reconnect` was renamed to `__should_resume` because that was what it actually did.
Line 325 was changed because I missed that in the docs when I read it the first time.
 ### Check off the following

### DIFF
--- a/pincer/core/gateway.py
+++ b/pincer/core/gateway.py
@@ -259,7 +259,6 @@ class Gateway:
             "%s Disconnected from Gateway due without any errors. Reconnecting.",
             self.shard_key
         )
-        self.__should_resume = True
         # ensure_future prevents a stack overflow
         ensure_future(self.start_loop())
 
@@ -312,8 +311,9 @@ class Gateway:
             self.shard_key
         )
 
-        await self.__socket.close()
         self.__should_resume = True
+        self.stop_heartbeat()
+        await self.__socket.close()
 
     async def handle_invalid_session(self, payload: GatewayDispatch):
         """|coro|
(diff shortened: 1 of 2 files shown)


pincer/core/gateway.py
```diff
@@ -310,19 +311,22 @@ async def handle_reconnect(self, payload: GatewayDispatch):
...
         when we tried to reconnect.
         """
-        _log.debug("%s Invalid session, attempting to relog...", self.shard_key)
-        self.__should_reconnect = False
-        await self.start_loop()
+
+        if payload.data:
+            _log.debug("%s Invalid session, attempting to reconnect...", self.shard_key)
+        else:
+            _log.debug("%s Invalid session, attempting to relog...", self.shard_key)
```
trag1c, line 327: eh? not the most readable thing but feels less redundant
```suggestion
        _log.debug(
            "%s Invalid session, attempting to re%s...",
            self.shard_key, ("log", "connect")[bool(payload.data)]
        )
```

trag1c (approved):

pincer/core/gateway.py
Sigmanificient: i'd prefer a `a if b else c` ternary instead of `(c, a)[b]`

Lunarmagpie: I can switch it to a ternary. It will be the entire word though cause splitting a word in half is kinda sus.

Sigmanificient (approved):

pincer/core/gateway.py
```diff
@@ -310,19 +311,23 @@ async def handle_reconnect(self, payload: GatewayDispatch):
...
-        self.__should_reconnect = False
-        await self.start_loop()
+
+        _log.debug(
+            "%s Invalid session, attempting to %s...",
+            self.shard_key,
+            "reconnect" if payload.data else "relog"
+        )
+
+        self.__should_resume = payload.data
```
Endercheif, line 330: If this is typed as a Boolean we should keep it as such.
```suggestion
        self.__should_resume = bool(payload.data)
```

Lunarmagpie: It is a bool. I will update the typehint to reflect that.

PR #359, opened 2022-01-08
:sparkles: :art: Remove dataclass decorator for APIObject

beastmatser:
## Changes

So it seems every instance of APIObject needs the `@dataclass` decorator, this is very repetitive, so in `HTTPMeta` I automatically set the decorator on the class

 ### Check off the following


beastmatser:
I used an old branch 😢

Sigmanificient:
Hi, i'm sorry but this pr will not be merged as it come from an older branch, and trying to solve conflict could lead to breaking a lot of features.

It would be really appreciated if you could redo your change in a up to date version and open a new pr for it.

VincentRPS:
Also in your new pull please don’t lint

PR #364, opened 2022-01-10
✨ Adding approx member count

Sigmanificient:
> before
image

```py
g = await self.get_guild(..., with_count=True)
member_count = g.approximate_member_count

await c.send(
    Embed(description=f"There is {member_count} members in this server!")
)
```
image

### DIFF
--- a/pincer/objects/guild/guild.py
+++ b/pincer/objects/guild/guild.py
@@ -434,7 +434,7 @@ class Guild(APIObject):
         data = await client.http.get(
             f"/guilds/{_id}",
             # Yarl don't support boolean params
-            params={"with_counts": ("true" * with_counts) or None},
+            params={"with_counts": "true" if with_counts else None},
         )
         channel_data = await client.http.get(f"/guilds/{_id}/channels")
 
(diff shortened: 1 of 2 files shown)


pincer/objects/guild/guild.py
```diff
@@ -426,7 +431,11 @@ async def from_id(cls, client: Client, _id: Union[int, Snowflake]) -> Guild:
         :class:`~pincer.objects.guild.guild.Guild`
             The new guild object.
         """
-        data = await client.http.get(f"/guilds/{_id}")
+        data = await client.http.get(
+            f"/guilds/{_id}",
+            # Yarl don't support boolean params
+            params={"with_counts": ("true" * with_counts) or None},
+        )
```
Lunarmagpie, line 438:
```suggestion
        data = await client.http.get(
            f"/guilds/{_id}",
            # Yarl don't support boolean params
            params={"with_counts": "true" if with_counts else None},
        )
```

Lunarmagpie (changes requested):

PR #376, opened 2022-01-10
:children_crossing: Add check on MinValue and MaxValue

beastmatser:
### Changes
A small and simple change: It adds value and type checking for the Min and MaxValue modifiers.

Newer users could potentially benefit from these preventive checks.

 ### Check off the following


beastmatser:
Turns out there were some misleading comments in the docstrings, I'll close this.

Issue #197, opened 2021-11-14
Feature: Easier permissions interaction


Skelmis:
**Is your feature request related to a problem? Please describe.**
Currently Pincer returns a ``str`` for permissions. 

**Describe the solution you'd like**
It would be nice to provide some form of interaction class with properties for each permission.

**Describe alternatives you've considered**
Doing the conversion and bitwise myself. Although most people won't even understand what bitwise is.

PR #378, opened 2022-01-10
:sparkles: Add permission handler

beastmatser:
### Changes

This adds a permission handler which can convert integers in the correct permission and vice versa.

 ### Check off the following


beastmatser:
#197

pincer/commands/arg_types.py
```diff
@@ -155,8 +155,6 @@ class ChannelTypes(Modifier):
```
Lunarmagpie, line 155: This file should not be modified

pincer/objects/guild/permissions.py
```diff
@@ -63,6 +66,15 @@ def __setattr__(self, name: str, value: Optional[bool]) -> None:
...
 
+    def __eq__(self, object) -> bool:
+        """
+        Permission equality is determined by comparing the integer values of the permissions
+        """
+        if isinstance(object, Permission):
+            return self.to_int() == object.to_int()
+        elif isinstance(object, tuple):
+            return self.to_int() == object
+
```
Sigmanificient, line 77:
```suggestion
    def __eq__(self, object) -> bool:
        """
        Permission equality is determined by comparing the integer values of the permissions
        """
        if isinstance(object, Permission):
            return self.to_int() == object.to_int()
        elif isinstance(object, tuple):
            return self.to_int() == object
        return False
```

Endercheif (changes requested):

pincer/objects/guild/permissions.py
```diff
@@ -101,3 +101,21 @@ def to_int(self) -> Tuple[int]:
                 deny |= enum.value
 
         return allow, deny
+
```
Sigmanificient:
```suggestion

```

pincer/objects/guild/permissions.py
```diff
@@ -101,3 +101,21 @@ def to_int(self) -> Tuple[int]:
...
+            if getattr(self, enum.name.lower()):
+                allow |= enum.value
+                
+        return allow
+    
+    @property
+    def deny(self) -> int:
+        deny = 0
+        for enum in PermissionEnums:
+            if getattr(self, enum.name.lower()) is False:
```
Sigmanificient, line 118:
```suggestion
            if not getattr(self, enum.name.lower()):
```

pincer/objects/guild/permissions.py
```diff
@@ -101,3 +101,21 @@ def to_int(self) -> Tuple[int]:
                 deny |= enum.value
 
         return allow, deny
+    
+    @property
```
Sigmanificient, line 105:
```suggestion
        return allow, deny

    @property
```

Sigmanificient (changes requested):

pincer/objects/guild/permissions.py
beastmatser: Since a permission can be None, this does not work, since `not None` is `True`

Lunarmagpie:
Looking over this, I think an implementation similar to this would be better https://github.com/Pincer-org/Pincer/blob/main/pincer/objects/events/presence.py#L134

beastmatser:
> ~Looking over this, I think an implementation similar to this would be better https://github.com/Pincer-org/Pincer/blob/main/pincer/objects/events/presence.py#L134.~ edit: Actually that doesn't work with None so that has to be ruled out

I feel like the whole purpose of this class is to ease the use of permissions, wouldn't a similar implementation to  [this](https://github.com/Pincer-org/Pincer/blob/main/pincer/objects/events/presence.py#L134) kind of defeat the purpose of this class

Lunarmagpie (changes requested):

Lunarmagpie:
Ignore what I said about |=

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,121 @@
...
+
+class Permission:
+    def __init__(self, **kwargs: Optional[bool]) -> None:
+        valid_perms = (enum.name.lower() for enum in PermissionEnums)
+        for perm in valid_perms:
+            setattr(self, perm, kwargs.pop(perm, None))
+
+        if kwargs:
+            invalid_perms = ', '.join(kwargs.keys())
+            raise ValueError(f"Invalid permissions were passed in: {invalid_perms}")
```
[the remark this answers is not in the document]
beastmatser: I already tried this but it interfered with `.from_int` because it would default everything to None, but I suppose I can give it another try

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,156 @@
+# Copyright Pincer 2021-Present
+# Full MIT License can be found in `LICENSE` at the project root.
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
```
Sigmanificient:
```suggestion

```

Sigmanificient (changes requested):

Sigmanificient (approved):

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+    create_private_threads: Optional[bool] = None
+    use_external_stickers: Optional[bool] = None
+    send_messages_in_threads: Optional[bool] = None
+    start_embedded_activities: Optional[bool] = None
+    moderate_members: Optional[bool] = None
+
+    def __setattr__(self, name: str, value: Optional[bool]) -> None:
+        if not isinstance(value, bool) and value is not None:
+            raise ValueError(f"Permission {name!r} must be a boolean or None")
+        return super().__setattr__(name, value)
```
Lunarmagpie, line 194: I don't think this is needed.

Lunarmagpie (approved):

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+    MODERATE_MEMBERS = 1 << 40
+
+
+@dataclass
+class Permission:
+    """
+    Allows for easier access to the permissions
+
+    Parameters
+    __________
```
Lunarmagpie, line 64:
```suggestion
    --------------
```

Lunarmagpie (changes requested):

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@dataclass
+class Permission:
+    """
+    Allows for easier access to the permissions
+
+    Parameters
+    --------------
+    create_instant_invite: :class:`Optional[bool]`
```
trag1c, line 65: I'm pretty sure it's ``Optional[:class:`bool`]``, at least that's what it's like in most of the codebase. If there's no difference then resolve this lol.

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+    MODERATE_MEMBERS = 1 << 40
+
+
+@dataclass
+class Permission:
+    """
+    Allows for easier access to the permissions
+
+    Parameters
+    --------------
```
trag1c, line 64:
```suggestion
    ----------
```

pincer/objects/guild/overwrite.py
```diff
@@ -31,3 +32,8 @@ class Overwrite(APIObject):
     type: int
     allow: str
     deny: str
+
+    @property
+    def permissions(self) -> Permission:
```
trag1c, line 37: the name's permission__s__ yet the return type is `Permission`, sounds a bit confusing imo

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+
+        return False
+
+    @classmethod
+    def from_int(cls, allow: int, deny: int) -> Permission:
+        """
+        Create a Permission object from an integer representation of the permissions (deny and allow)
+
+        Parameters
+        __________
```
trag1c, line 213:
```suggestion
        ----------
```

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        clsobj = cls()
+
+        for enum in Permissions:
+            if bool(enum.value & allow):
+                setattr(clsobj, enum.name.lower(), True)
+            elif bool(enum.value & deny):
+                setattr(clsobj, enum.name.lower(), False)
+            else:
+                setattr(clsobj, enum.name.lower(), None)
```
trag1c, line 227:
```suggestion
        for enum in Permissions:
            if enum.value & allow:
                setattr(clsobj, enum.name.lower(), True)
            elif enum.value & deny:
                setattr(clsobj, enum.name.lower(), False)
            else:
                setattr(clsobj, enum.name.lower(), None)
```
Or even this (the parentheses at L225 ain't needed but make it slightly more readable)
```suggestion
        for enum in Permissions:
            setattr(
                clsobj, enum.name.lower(),
                True if enum.value & allow else
                (False if enum.value & deny else None)
            )
```

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+
+        return clsobj
+
+    def to_int(self) -> Tuple[int]:
+        """
+        Convert the Permission object to an integer representation of the permissions (deny and allow)
+
+        Returns
+        _______
```
trag1c, line 236:
```suggestion
        -------
```

trag1c (changes requested):

pincer/objects/guild/permissions.py
Sigmanificient: Maybe not the shortest but way more readable

```suggestion
        for enum in Permissions:
            value = None

            if enum.value & allow:
                value = True

            elif enum.value & deny:
                value = False

            setattr(clsobj, enum.name.lower(), value)
```

Endercheif (changes requested):
I agree with trag1c

Endercheif (approved):

Sigmanificient (approved):

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+        return super().__setattr__(name, value)
+
+    def __eq__(self, object) -> bool:
+        """
+        Permission equality is determined by comparing the integer values of the permissions
+        """
+        if isinstance(object, Permission):
+            return self.to_int() == object.to_int()
+        elif isinstance(object, tuple):
+            return self.to_int() == object
```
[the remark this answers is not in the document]
Lunarmagpie: The function returns a tuple of ints. Maybe the function name should be changed because I was confused by this too.

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,272 @@
...
+        _______
+        :class:`Tuple[int]`
+            The integer representation of the permissions that are allowed and denied
+        """
+        allow = 0
+        deny = 0
+        for enum in Permissions:
+            if getattr(self, enum.name.lower()):
+                allow |= enum.value
+            elif getattr(self, enum.name.lower()) is False:
```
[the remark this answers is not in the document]
Lunarmagpie: There's different behavior it the attribute is None or false

Lunarmagpie (approved):

pincer/objects/guild/permissions.py
```diff
@@ -0,0 +1,262 @@
...
+# Full MIT License can be found in `LICENSE` at the project root.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Tuple, Optional
+
+
+class PermissionEnum(Enum):
```
trag1c, line 11: maybe?
```suggestion
from enum import IntEnum
from typing import Tuple, Optional


class PermissionEnum(IntEnum):
```

trag1c (approved):

pincer/objects/guild/permissions.py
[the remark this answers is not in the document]
beastmatser: A few ideas:
- to_ints
- to_tuple
- convert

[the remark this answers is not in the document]
Lunarmagpie: I think you should do `to_ints` since it matches with `from_ints`

My other idea is `to_permission_tuple` but thats kinda long.

Sigmanificient (approved):

Lunarmagpie (approved):

PR #380, opened 2022-01-10
:sparkles: Adding message history

Sigmanificient:
```py
c = await self.get_channel(...)

async for m in c.history(limit=1000):
    # To your stuff
```

### DIFF
--- a/pincer/objects/guild/channel.py
+++ b/pincer/objects/guild/channel.py
@@ -805,7 +805,6 @@ class TextChannel(Channel):
             An iterator of messages.
         """
 
-
         if limit is None:
             limit = float('inf')
 
@@ -825,13 +824,10 @@ class TextChannel(Channel):
             if not raw_messages:
                 break
 
-            for _message in raw_messages[:-1]:
+            for _message in raw_messages:
                 yield UserMessage.from_dict(_message)
 
-            last = UserMessage.from_dict(raw_messages[-1])
-            yield last
-
-            before = last.id
+            before = raw_messages[-1]['id']
             limit -= retrieve
 
 
--- a/pincer/objects/guild/channel.py
+++ b/pincer/objects/guild/channel.py
@@ -790,13 +790,13 @@ class TextChannel(Channel):
 
         Parameters
         ----------
-        around
+        around : Optional[Union[:class:`int`, :class:`Snowflake`]]
             The message ID to look around.
-        after
+        after : Optional[Union[:class:`int`, :class:`Snowflake`]]
             The message ID to look after.
-        before
+        before : Optional[Union[:class:`int`, :class:`Snowflake`]]
             The message ID to look before.
-        limit
+        limit : Optional[Union[:class:`int`, :class:`Snowflake`]]
             The maximum number of messages to return.
 
         Returns
--- a/pincer/objects/guild/channel.py
+++ b/pincer/objects/guild/channel.py
@@ -805,14 +805,6 @@ class TextChannel(Channel):
             An iterator of messages.
         """
 
-        if isinstance(before, int):
-            before = Snowflake(before)
-
-        if isinstance(after, int):
-            after = Snowflake(after)
-
-        if isinstance(around, int):
-            around = Snowflake(around)
 
         if limit is None:
             limit = float('inf')
--- a/pincer/objects/guild/channel.py
+++ b/pincer/objects/guild/channel.py
@@ -779,6 +779,69 @@ class TextChannel(Channel):
             await self._http.get(f"channels/{self.id}/messages/{message_id}")
         )
 
+    async def history(
+        self, limit: int = 50,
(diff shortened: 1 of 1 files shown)


pincer/objects/guild/channel.py
```diff
@@ -779,6 +779,69 @@ async def fetch_message(self, message_id: int) -> UserMessage:
...
+            after = Snowflake(after)
+
+        if isinstance(around, int):
+            around = Snowflake(around)
+
+        if limit is None:
+            limit = float('inf')
+
+        while limit > 0:
+            retrieve = min(limit, 100)
```
Lunarmagpie, line 821: limit HAS to be capped to 100. If they choose a number too big there will be rate limits and that will cause  it to stall because of an "invisible error"

Lunarmagpie (changes requested):

pincer/objects/guild/channel.py
Sigmanificient: I don't agree, we need to provide an user-friendly way to get all messages within a channel, even if it is extremely slow. 
Discord.py do it has well, see `channel.history(limit=...).flatten()`

Also it deal very nicely with ratelimiting.
I test if with 4000 messages, in `41.18647676300316 seconds`

Lunarmagpie: I see your point but you need to change things
1. Limit should never be set to infinity if the user doesn't explicitly do so. I would make it 100 to prevent rate limits from a mistake.
2. There can be a lot of api calls so you should use gather. You can even use itertools.chain around gather. I bet that would be useful here.

pincer/objects/guild/channel.py
```diff
@@ -779,6 +779,69 @@ async def fetch_message(self, message_id: int) -> UserMessage:
...
+                    'before': before,
+                    'after': after,
+                    'around': around,
+                }
+            )
+
+            if not raw_messages:
+                break
+
+            for _message in raw_messages[:-1]:
```
Lunarmagpie, line 836: Sorry on mobile
```python
before = raw_messages[-1].I'd
for message in raw_messages:
    yield UserMessage.from_dict(message)
```

Lunarmagpie (changes requested):

pincer/objects/guild/channel.py
Sigmanificient: `raw_messages` is not yet converted payload, so you cant access `.id`, i guess we can do `["id"]` tho

Lunarmagpie: I think that would be bettr

pincer/objects/guild/channel.py
```diff
@@ -779,6 +779,69 @@ async def fetch_message(self, message_id: int) -> UserMessage:
             await self._http.get(f"channels/{self.id}/messages/{message_id}")
         )
 
+    async def history(
+        self, limit: int = 50,
+        before: Union[int, Snowflake] = None,
+        after: Union[int, Snowflake] = None,
+        around: Union[int, Snowflake] = None,
```
Endercheif, line 786:
```suggestion
        before: Optional[Union[int, Snowflake]] = None,
        after: Optional[Union[int, Snowflake]] = None,
        around: Optional[Union[int, Snowflake]] = None,
```

pincer/objects/guild/channel.py
```diff
@@ -779,6 +779,69 @@ async def fetch_message(self, message_id: int) -> UserMessage:
...
+        """
+
+        if isinstance(before, int):
+            before = Snowflake(before)
+
+        if isinstance(after, int):
+            after = Snowflake(after)
+
+        if isinstance(around, int):
+            around = Snowflake(around)
```
Endercheif, line 815: Is this needed? Functionally, an `int` and a `Snowflake` act the same other than a few added methods, none of which that are used here.
```suggestion
```

Endercheif (changes requested):

Lunarmagpie (approved):

pincer/objects/guild/channel.py
```diff
@@ -779,6 +779,57 @@ async def fetch_message(self, message_id: int) -> UserMessage:
...
+        Parameters
+        ----------
+        around
+            The message ID to look around.
+        after
+            The message ID to look after.
+        before
+            The message ID to look before.
+        limit
+            The maximum number of messages to return.
```
Endercheif, line 800: Add types in the docstring
```suggestion
        around : Optional[Union[:class:`int`, :class:`Snowflake`]]
            The message ID to look around.
        after : Optional[Union[:class:`int`, :class:`Snowflake`]]
            The message ID to look after.
        before : Optional[Union[:class:`int`, :class:`Snowflake`]]
            The message ID to look before.
        limit : Optional[Union[:class:`int`, :class:`Snowflake`]]
            The maximum number of messages to return.
```

Endercheif (approved):

Issue #356, opened 2022-01-07
:sparkles: Add voice states for the client


Lunarmagpie:
The should be a function to request a `VoiceState` and `VoiceServer` object from the gateway.
The function should look like this (this is inspired by Hikari)
```python
@command
async def something(self: Client):
    self.voice.connect_to(GUILD_ID, channel_id, connection_type=InheritsFromVoiceABC)
```
There should also be a leave function that can be run when the bot is shut down. We can make that run automatically later.
A `VoiceABC` object will be added so that people can make their own voice systems that work with Pincer easily.

Some Technical Notes:
A system to save a reference to each shard needs to be added. There then needs to be a way to retrieve them using a guild id. https://discord.com/developers/docs/topics/gateway#sharding-sharding-formula May be a good idea to implement that in a PR before touching this.

This supersedes #217

Arthurdw:
Closed due to depreciation, thank you for your effort ♥

PR #384, opened 2022-01-12
:sparkles: added get_shard method to client

Lunarmagpie:
### Changes

-   `adds`: get_shard method to client
-   `fixed`: get_shard method to client
-   `improvements`: get_shard method to client

This is required for implementing good voice support so its part of #356
I've decided to stop making PRs that are too big :smiley: 

 ### Check off the following


pincer/client.py
```diff
@@ -562,8 +566,38 @@ async def start_shard(self, shard: int, num_shards: int):
...
+
+        Parameters
+        ----------
+        guild_id : Optional[~pincer.utils.snowflake.Snowflake]
+            The guild_id of the shard to look for
+        num_shards : Optional[int]
+            The number of shards.
+        """
+        if not guild_id:
+            return self.shards[0]
```
Endercheif, line 592: On L594 there is `if not self.shards`. This would imply a chance at the shards array being empty would it not? If so, this could throw an `IndexError`.

pincer/utils/__init__.py
```diff
@@ -23,10 +24,10 @@
...
 
 __all__ = (
-    "", "APINullable", "APIObject", "ChannelProperty",
+    "APINullable", "APIObject", "ChannelProperty",
     "CheckFunction", "Color", "Coro", "EventMgr", "GuildProperty", "MISSING",
     "MissingType", "Snowflake", "Task", "TaskScheduler", "Timestamp", "chdir",
     "choice_value_types", "get_index", "get_params",
     "get_signature_and_params", "remove_none", "replace", "should_pass_cls",
-    "should_pass_ctx"
+    "should_pass_ctx", "calculate_shard_id"
```
Endercheif, line 32: Alphabetical pls
```diff
+   "APINullable", "APIObject", "calculate_shard_id", "ChannelProperty",
    "CheckFunction", "Color", "Coro", "EventMgr", "GuildProperty", "MISSING",
    "MissingType", "Snowflake", "Task", "TaskScheduler", "Timestamp", "chdir",
    "choice_value_types", "get_index", "get_params",
    "get_signature_and_params", "remove_none", "replace", "should_pass_cls",
-   "should_pass_ctx", "calculate_shard_id"
+   "should_pass_ctx"
```

pincer/utils/shards.py
```diff
@@ -0,0 +1,14 @@
...
+
+def calculate_shard_id(guild_id, num_shards):
+    """Calculates the shard receiving the events for a specified guild
+
+    Parameters
+    ----------
+    guild_id : Optional[~pincer.utils.snowflake.Snowflake]
+        The guild_id of the shard to look for
+    num_shards : Optional[int]
+        The number of shards.
```
Endercheif, line 12: Why are these typed as `Optional[TYPE]`?

Endercheif (changes requested):

pincer/utils/__init__.py
Sigmanificient: We have a worflow that will do it automagically

pincer/utils/shards.py
```diff
@@ -0,0 +1,14 @@
+# Copyright Pincer 2021-Present
+# Full MIT License can be found in `LICENSE` at the project root.
+
+def calculate_shard_id(guild_id, num_shards):
```
Sigmanificient, line 4: *May need to import Snowflake*
```suggestion
def calculate_shard_id(guild_id: Union[Snowflake, int], num_shards: int) -> int:
```

pincer/utils/__init__.py
Endercheif: I just noticed the `sorted`. Thanks

pincer/utils/shards.py
Lunarmagpie: No guild id = get dm guild. This means that you can pass in ctx.guild_id if all interactions and it will get the correct guild if it is a dm.
No num_shards = default to most recent shard count. Just makes things easier to write.
I will update docs when i get home

Issue #375, opened 2022-01-10
:bug: Workflow "sort-all" fail when nothing to commit


Sigmanificient:
### Describe the bug

Since sort-all is active on all branch, it now fail on pr if there is nothing to sort

### Relevant console output

see [here](https://github.com/Pincer-org/Pincer/runs/4761952669?check_suite_focus=true)

Lunarmagpie:
I'm pretty sure this works now

Sigmanificient:
yeah

PR #385, opened 2022-01-12
:construction_worker: Fix license generator

beastmatser:
### Changes

The workflow doesn't fail if there are no changes to commit, as you can see the license was added this time.

 ### Check off the following

-   [ ] My Code follows the pep8 code style.


Lunarmagpie (approved):

Sigmanificient:
@beastmatser Could you help us also to fix a similar issue [##375](https://github.com/Pincer-org/Pincer/issues/375)?
It would be epic!

PR #403, opened 2022-01-22
♻️ Using a Custom class for Generator to be awaitable

Sigmanificient:
### Changes

-   `adds`: a Custom class for Generator to be awaitable

 ### Check off the following

-   [X] I have tested my changes with the current requirements
-   [X] My Code follows the pep8 code style.

### DIFF
--- a/pincer/utils/api_data.py
+++ b/pincer/utils/api_data.py
@@ -1,3 +1,6 @@
+# Copyright Pincer 2021-Present
+# Full MIT License can be found in `LICENSE` at the project root.
+
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


pincer/utils/api_data.py
```diff
@@ -0,0 +1,36 @@
+# Copyright Pincer 2021-Present
+# Full MIT License can be found in `LICENSE` at the project root.
+
+from __future__ import annotations
```
trag1c, line 4: Not sure where the import is used?

trag1c (approved):

Issue #425, opened 2022-02-06
:bug: commands parameters are always str


Kylianalex:
### Describe the bug

example:
adding a Role, int or Channel parameter to a command get a ``str`` object

### Steps to Reproduce

```markdown
1. create a command with a any parameter
2. print type(param)
```

### Pincer Version used

git

Kylianalex:
fixed, going to pull

Arthurdw:
Closed due to depreciation, thank you for your effort ♥

PR #438, opened 2022-03-03
🐛 Fixed command parameters types

Kylianalex:
was not working and everything was str
also bool("False") will return True because the string is not empty

### Changes

-   `improvements`: now a parameter with type int will return an int, not str

 ### Check off the following

### DIFF
--- a/pincer/objects/app/interactions.py
+++ b/pincer/objects/app/interactions.py
@@ -154,12 +154,15 @@ class Interaction(APIObject, ChannelProperty, GuildProperty):
             elif option.type == AppCommandOptionType.NUMBER:
                 option.value = float(option.value)
 
-            elif option.type == AppCommandOptionType.USER:
+            elif option.type is AppCommandOptionType.USER:
                 user = self.return_type(option, self.data.resolved.members)
-                user.set_user_data(
-                    self.return_type(option, self.data.resolved.users)
-                )
-                option.value = user
+                if user:
+                    user.set_user_data(
+                        self.return_type(option, self.data.resolved.users)
+                    )
+                    option.value = user
+                else:
+                    option.value = self.return_type(option, self.data.resolved.users)
 
             elif option.type == AppCommandOptionType.CHANNEL:
                 option.value = self.return_type(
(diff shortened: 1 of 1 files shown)


pincer/objects/app/interactions.py
```diff
@@ -145,33 +145,36 @@ def __post_init__(self):
...
+            if option.type == AppCommandOptionType.STRING:
                 option.value = str(option.value)
-            elif option.type is AppCommandOptionType.INTEGER:
+            elif option.type == AppCommandOptionType.INTEGER:
                 option.value = int(option.value)
-            elif option.type is AppCommandOptionType.BOOLEAN:
-                option.value = bool(option.value)
-            elif option.type is AppCommandOptionType.NUMBER:
+            elif option.type == AppCommandOptionType.BOOLEAN:
+                option.value = option.value == "True"
```
Sigmanificient, line 153: I'm not sure of this change

pincer/objects/app/interactions.py
```diff
@@ -145,33 +145,36 @@ def __post_init__(self):
             return
 
         for option in self.data.options:
-            if option.type is AppCommandOptionType.STRING:
+            if option.type == AppCommandOptionType.STRING:
```
Sigmanificient, line 148: @Lunarmagpie not sure, but this is alright?

pincer/objects/app/interactions.py
Kylianalex: the problem is that discord send "True" or "False" as str and i don't know how to make it boolean with the pep8 style

pincer/objects/app/interactions.py
Kylianalex: #425 the "is" statement is not working for me in python 3.9

Arthurdw: Okay, I'm assuming that our APIObject isn't properly converting the socket payload to an object. As I'm not up to date with the current state of the way the lib handles this, I can't provide a solution rn.  
  
One thing I do know is that this is not the expected behavior, and we should model around the expected behavior _(which is that there is an object, so a `is`_ statement should be used_). Nonetheless,_ this issue should be addressed.

PR #469, opened 2022-05-28
Nox stuff

Endercheif:
### Changes

It has come to my attention that many of the measure we have in place are not being run or used as much. Nox will fix that. 
Before doing a pr, on can run `nox` on their local machine to run tests and lints. As a workflow, they will also run.

 ### Check off the following


pyproject.toml
```diff
@@ -74,5 +75,5 @@ warn_unused_configs = true
 no_implicit_reexport = true
 
 [tool.black]
-line-length = 80
+line-length =  88
```
Arthurdw, line 78: Any reason for this change?

Endercheif: In .flake8 and .editorconfig it is set to 88

Sigmanificient:
What is tox doing?

Endercheif:
> What is tox doing?

tox (thing I removed) did nothing because it was never integrated into the code base, it was an extra dependancy. nox (what I added) runs black, isort, flake8, pytest, mypy all in one command. Prior to this, it was clear many had not ran these individually as they were supposed to. nox makes it easy and will also auto run as a GitHub workflow

Sigmanificient:
> > What is tox doing?
> 
> tox (thing I removed) did nothing because it was never integrated into the code base, it was an extra dependancy. nox (what I added) runs black, isort, flake8, pytest, mypy all in one command. Prior to this, it was clear many had not ran these individually as they were supposed to. nox makes it easy and will also auto run as a GitHub workflow

We maybe dont want to run mypy cause there is a lot of type errors currently

trag1c (approved):

Sigmanificient (approved):

Sigmanificient:
nox still failling xD

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+      matrix:
+        os: [ ubuntu, windows, macos ]
+
+    name: Nox on ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Python Setup
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
Lunarmagpie, line 19: Nox should be run on all supported versions of python. You should use the matrix thing. Heres an example from one of my projects https://github.com/magpie-dev/hikari-crescent/blob/main/.github/workflows/pytest.yml

noxfile.py
```diff
@@ -0,0 +1,27 @@
...
+@nox.session
+def lint(session: nox.Session):
+    session.install("flake8")
+    session.run("flake8", "pincer")
+
+
+@nox.session
+def mypy(session: nox.Session):
+    session.install("mypy")
+    session.run("mypy", "pincer")
```
Lunarmagpie, line 27: Mypy shouldn't be run. Theres far too many type issues so its always gonna fail.

Lunarmagpie (changes requested):

noxfile.py
Sigmanificient: Yes that what i was saying too, we should run it later

Sigmanificient, line 27:
```suggestion
"""
Mypy will be be ran within a workflow until pincer provide complete typing
that will not fail for every push.

@nox.session
def mypy(session: nox.Session):
   session.install("mypy")
   session.run("mypy", "pincer")
"""
```

pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
 "Discord" = "https://discord.gg/8WkYz3fNFm"
 
 [tool.poetry.dependencies]
-python = ">=3.8"
+python = ">=3.8,<3.11"
```
Sigmanificient, line 28: Breaking changes on python 3.11?

Sigmanificient (changes requested):

pyproject.toml
Arthurdw: I don't think there are breaking changes for us.

Sigmanificient: Why changing this then?

.github/workflows/ci.yml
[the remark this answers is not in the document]
Sigmanificient: @Endercheif

noxfile.py
Sigmanificient: @Endercheif

pyproject.toml
Endercheif: @Sigmanificient @Arthurdw Poetry is very picky about supported version of dependencies. One such dependency (Nox I believe) does not support anything greater than 3.11 so the install fails. Unless this is added.

Sigmanificient:
Can we just format the coe with black one and for all?

Endercheif:
The only reason the Nox is failing, is because the code is not formatted correctly. 
We have two options. 
- Merge this now and in a different PR resolve the formatting issue or 
- In this PR, format it.

I would say the second one just to keep the PRs organized.

Sigmanificient (approved):

Arthurdw (approved):
Seems okay to me.

